PR #2, opened 2015-03-25
Improve description of -v option

lencioni:
I found the description for this option to be confusing, partly because
I don't think "repeate" is actually a word, and partly because if it
was a typo for "repeat", it hardly makes sense to me here. In any case,
these words seem unnecessary and can probably just be removed entirely.

I also remove the two occurrences of trailing whitespace in the readme.


fkling:
Thanks! You are right. There reason for that is that I used an other arguments parser before, which allowed to do `-vv` or `-vvv`, but the new one doesn't and I forgot to update the description.

PR #3, opened 2015-03-31
Change implementation of collections

fkling:
Note: This is not a breaking change. It only changes how collections are
represented internally and solves a couple of issues with the current
implementation.

#### How typed collections currently work

ast-types defines a type hierarchy. For example, a "BlockStatement" has this
hierarchy:

    BlockStatement <- Statement <- Node <- Printable

I.e. a "BlockStatement" is also a "Statement", a "Node", etc.

jscodeshift uses this hierarchy to model collection classes. Every collection
class inherits from its supertype/parent collection class. For example,
"BlockStatementCollection" inherits from a "StatementCollection",
etc.

Whenever `registerMethods(methods, type)` is called, classes for "type"
and all its supertypes are created, if they don't exist yet, and are extended
with "methods". This works (obviously) but has some disadvantages. The biggest
are:

**Unable to account for multiple parents**

Some types actually have multiple parent types. For example, a
"FunctionExpression" is both a "Function" and an "Expression". The
following currently fails because "ExpressionCollection" is not in the
inheritance chain of "FunctionExpressionCollection":

```js
jscodeshift.registerMethods({
  expressionMethod: function() {}
}, jscodeshift.Expression);

jscodeshift('(function foo() {})')
  .find(jscodeshift.FunctionExpression)
  .expressionMethod();
```

This will throw the unhelpful error message: "undefined is not a
function".

Which brings me to the next issue:

**Unhelpful error messages**

If you are getting a collection of a different type (A) than you expect (say B),
and trying to call a method specific only to B, then you will get the
error message "undefined is not a function". At this point, it's not
clear whether the issue is just a typo, different naming of the method
or if you get a differently typed collection.

#### How typed collections work after this change

Instead of dynamically creating classes which inherit from each other,
there will only be a single `Collection` class. Each collection instance
has an internal `_types` property which holds a list of types the
collection supports (i.e. all nodes in the collection are of these types).

Registered methods are just added to the prototype of the `Collection` class.
If such a method is called, we will first check whether the type it was
registered for is in the internal list of types. If it is not, we can
provide a helpful error, containing which type the method was registered
for and which types the current collection is made of.

Since everything is flat now, there is also no problem with multiple
parents anymore. The above example would just work.

This implementation currently does not allow to register a method with
the same name for different types. I think this is fine, because having
a method that does different things depending on the type of the
collection can be confusing. However, if it turns our that this is too
restrictive we can consider alternatives.

---

After this change we should also think about restructuring the source
code to better reflect the new structure.

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -14,43 +14,10 @@ var assert = require('assert');
 var recast = require('recast');
 var _ = require('lodash');
 
-var types = recast.types;
-var NodePath = types.NodePath;
-var Node = types.namedTypes.Node;
-
-
-var _typedCollectionCache = Object.create(null);
-
-function _createCollectionForType(type) {
-  /*jshint evil:true*/
-  // Yeah this is ugly, but I want to give the function a proper name to make
-  // debugging easier. Maybe can omit the name at some point.
-  var Constr = eval(
-    '(function ' + type + 'Collection() { Collection.apply(this, arguments);})'
-  );
-
-  // TODO: Try to find a solution for types having multiple supertypes.
-  // E.g. a FunctionExpression has the supertypes Function and Expression.
-  // Not sure if this is an issue so far.
-  var supertypes = types.getSupertypeNames(type.toString());
-  var superTypePrototype = supertypes[0] ?
-    _getCollectionForType(supertypes[0]).prototype :
-    Collection.prototype;
-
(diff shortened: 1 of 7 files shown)


cpojer:
:+1:

fkling:
It seems the tests in npm-react-codemod still pass with these changes. Will merge tomorrow.

PR #6, opened 2015-04-04
Rename --verbosity option to --verbose

wincent:
For consistency with most other command-line tools out there. Note that
this is a breaking change, but I think we can probably get away without
adding a separate option and keeping `--verbosity` around as a duplicate
(which would clutter the usage output).


fkling:
That seems reasonable. Maybe we can push that to a new release branch, which will also include things like the config file and preprocess stuff.

fkling:
Merged into v0.2 branch with f605b785cb359a2a087f1ad3926ebf2bda1295a9

PR #8, opened 2015-04-04
Replace references to docgen in the test suite

wincent:


fkling:
Thanks! I guess that shows where I copied that code from.

PR #15, opened 2015-07-03
use babel parser for recast parsing - fixes #14

### LINKED ISSUE #14: Use Babel's parser instead of recast's default parser (esprima-fb)

hzoo:
#14 

I ~~can* also~~ added another commit to also use babel for build/watch and remove react-tools

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "bugs": "https://github.com/facebook/jscodeshift/issues",
   "main": "index.js",
   "scripts": {
-    "build": "rm -rf dist; jsx --harmony src/ dist/",
-    "watch": "jsx --harmony src/ dist/ -w",
+    "build": "rm -rf dist; babel src --out-dir dist",
+    "watch": "babel src --out-dir dist -w",
     "test": "npm run build && jest",
     "prepublish": "npm run build"
   },
@@ -32,9 +32,9 @@
     "recast": "^0.10.10"
   },
   "devDependencies": {
+    "babel": "^5.6.14",
     "es6-promise": "^2.0.1",
     "jest-cli": "^0.4.0",
-    "react-tools": "^0.13.1",
     "temp": "^0.8.1"
   },
   "jest": {
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "async": "^0.9.0",
+    "babel-core": "^5.6.15",
     "cli-color": "^0.3.2",
-    "esprima-fb": "^14001.1.0-dev-harmony-fb",
     "lodash": "^3.5.0",
     "nomnom": "^1.8.1",
     "recast": "^0.10.10"
--- a/scripts/test-preprocess.js
+++ b/scripts/test-preprocess.js
@@ -10,10 +10,10 @@
 
 'use strict';
 
-var reactTools = require('react-tools');
+var babel = require('babel-core');
 
 function process(source) {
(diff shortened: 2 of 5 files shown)


hzoo:
I guess the issue is with jest/jsdom/contextify? https://github.com/facebook/jest/issues/243 - tests pass on node 0.10

cpojer:
Awesome.

We should probably upgrade to io.js.

@fkling now that babel has become the de-facto standard, should we just run the transform file through babel as well? Best case users of jscodeshift don't need their own wrapper library (like I build), worst case we transform a file twice, which shouldn't be a big deal.

benjamn:
+1 to switching to the Babel parser for everything. Recast can parse/visit/print all the types it uses.

fkling:
@cpojer: Sure. Maybe we can add a flag that disables transpilation (I *think* Babel has (had) some issues with double compilation).
@hzoo: can you have a look why Travis fails? Do the tests pass locally?

hzoo:
@fkling Yeah made a comment above

> I guess the issue is with jest/jsdom/contextify? facebook/jest#243 - tests pass on node 0.10

It's working locally for me on node v0.10.38 (ubuntu).

fkling:
Yeah, I missed that comment... strange though, because Travis also fails for v0.10, which seems to indicate that it is not a jest problem?

hzoo:
Ok I just cleared my npm cache and did npm install again and have 3 local tests failing too now...

hzoo:
Weird getting a syntax error

```text
FAIL  src/__tests__/core-test.js
SyntaxError: /home/henry/Dev/jscodeshift/src/__tests__/core-test.js:
/home/henry/Dev/jscodeshift/src/core.js:
/home/henry/Dev/jscodeshift/node_modules/babel-core/index.js:
... a lot more  babel files
/home/henry/Dev/jscodeshift/node_modules/babel-core/node_modules/is-integer/node_modules/is-nan/node_modules/define-properties/node_modules/object-keys/index.js:
Unexpected identifier
    at Contextify.sandbox.run (/home/henry/Dev/jscodeshift/node_modules/jest-cli/node_modules/jsdom/node_modules/contextify/lib/contextify.js:12:24)
    at JSDomEnvironment.runSourceText (/home/henry/Dev/jscodeshift/node_modules/jest-cli/src/JSDomEnvironment.js:111:22)
```

hzoo:
@fkling @benjamn Has no one used babel/acorn with recast yet? If not, then think it's because it's not compatible with esprima? In `babel-eslint` we've had to do make a https://github.com/babel/babel-eslint/blob/master/acorn-to-esprima.js. by convert acorn tokens to esprima somewhat compatible. Although I guess that wouldn't be why there's an syntax error - will check it out more later

cpojer:
Been trying to get this to work to no avail so far. Ugh.

hzoo:
:disappointed: I tried iojs@2.3.4 with nvm and facebook/jest#0.5.x and got the same kind of syntax error (and this is with no other changes like using babel)

```
 FAIL  src/collections/__tests__/JSXElement-test.js
SyntaxError: /home/henry/Dev/jscodeshift/src/collections/__tests__/JSXElement-test.js: Unexpected identifier
  at Object.exports.runInContext (vm.js:43:16)
  at JSDomEnvironment.runSourceText (/home/henry/Dev/jscodeshift/node_modules/jest-cli/src/JSDomEnvironment.js:110:8)
```

fkling:
Just to establish a baseline: I can run the tests on the master branch without problems on Node v0.10.36.

PR #16, opened 2015-07-11
[RFC] "Templates" -- first steps

fkling:
"Templates" may not be the right term for this, open to suggestions.

This adds three tagged templates to the core API:

- `jscodeshift.template.expression`
- `jscodeshift.template.statement`
- `jscodeshift.template.statements`

Their purpose is to make it easier to replace nodes with more complex constructs, without having to build the replacement AST yourself. For example, the following replaces all matching identifiers with a function
call:

```js
let {expression} = jscodeshift.template;

// replaces all `bar` identifiers with the function call `foo(bar)`
jscodeshift(src)
  .find('Identifier', {name: 'bar'})
  .replaceWith(path => expression`foo(${path.node})`);
```

The difference between the three function is just how the provided
code snippet is parsed:

- `expression` returns an expression node
- `statement` returns a statement node
- `statements` returns an array of statement nodes

It is now also possible to replace a node with multiple nodes. This is especially useful for replacing a single statement with multiple statements (still have to experiment with expression -> multiple expressions replacements).

---

Still needs some more tests, but putting it out here for comments.

---

**Note:** You can currently only inject the nodes in places where an identifier would be valid, i.e. this won't work:

```js
`switch(x) {
   ${foo}
   case bar:
     // ...
}
`
```

I was thinking about extending the acorn parser with a new token type and AST node type which would be valid everywhere and which would represent a "node placeholder". But I don't know yet if this is possible and how difficult it would be.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "async": "^0.9.0",
     "babel": "^5.6.14",
+    "babel-runtime": "^5.6.18",
     "cli-color": "^0.3.2",
     "esprima-fb": "^15001.1.0-dev-harmony-fb",
     "lodash": "^3.5.0",
@@ -39,10 +40,15 @@
     "temp": "^0.8.1"
   },
   "jest": {
-    "scriptPreprocessor": "./node_modules/babel-jest",
+    "scriptPreprocessor": "<rootDir>/node_modules/babel-jest",
+    "preprocessCachingDisabled": true,
     "testPathDirs": [
       "src",
       "bin"
+    ],
+    "unmockedModulePathPatterns": [
+      "babel-runtime",
+      "babel"
     ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
   "bugs": "https://github.com/facebook/jscodeshift/issues",
   "main": "index.js",
   "scripts": {
-    "build": "rm -rf dist; jsx --harmony src/ dist/",
-    "watch": "jsx --harmony src/ dist/ -w",
-    "test": "npm run build && jest",
+    "build": "rm -rf dist; babel src/ --out-dir dist/",
+    "watch": "babel src/ --out-dir dist/ --watch",
+    "test": "jest",
     "prepublish": "npm run build"
   },
   "bin": {
@@ -25,6 +25,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "async": "^0.9.0",
+    "babel": "^5.6.14",
     "cli-color": "^0.3.2",
     "esprima-fb": "^15001.1.0-dev-harmony-fb",
     "lodash": "^3.5.0",
@@ -32,13 +33,13 @@
     "recast": "^0.10.10"
   },
   "devDependencies": {
+    "babel-jest": "^5.3.0",
     "es6-promise": "^2.0.1",
     "jest-cli": "^0.4.0",
-    "react-tools": "^0.13.1",
     "temp": "^0.8.1"
   },
   "jest": {
-    "scriptPreprocessor": "./scripts/test-preprocess.js",
+    "scriptPreprocessor": "./node_modules/babel-jest",
     "testPathDirs": [
       "src",
       "bin"
--- a/src/__tests__/template-test.js
+++ b/src/__tests__/template-test.js
@@ -1,0 +1,180 @@
+/*global jest, defined, it, expect, beforeEach*/
+
+jest.autoMockOff();
+
+let jscodeshift = require('../core');
+
+describe('Templates', () => {
+  let statements;
+  let statement;
+  let expression;
+
+  beforeEach(() => {
+    ({expression, statement, statements} = require('../template'));
+  });
+
+  it('interpolates expression nodes with source code', () => {
+
+    let input =
+`var foo = bar;
+if(bar) {
+  console.log(42);
+}`;
+
+    let expected =
+`var foo = alert(bar);
+if(alert(bar)) {
+  console.log(42);
+}`;
+
+    expect(
+      jscodeshift(input)
+        .find('Identifier', {name: 'bar'})
+        .replaceWith(path => expression`alert(${path.node})`)
+        .toSource()
+    ).toEqual(expected);
+  });
+
+  it('interpolates statement nodes with source code', () => {
+    let input =
+`for (var i = 0; i < 10; i++) {
+  console.log(i);
+  console.log(i / 2);
+}`;
+
+    let expected =
+`var i = 0;
+
+while (i < 10) {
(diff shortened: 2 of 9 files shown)


cpojer:
Interesting. What I had on my mind was something like this:

```js
node.replaceWith(
  parse`var $foo = $bar($args)`,
  {
    foo: j.identifier('javascript'),
    bar: j.identifier('is'),
    args: [j.identifier('awesome')],
  }
);
```

and it would print to `var javascript = is(awesome);`

What do you think about this approach? I'm a little worried your idea, while sound in theory, makes the API more complex as it introduces more ways of creating nodes.

fkling:
Is the only difference that you are using placeholders instead of interpolation?

In that case, I don't really see how this idea is different from mine.



On Jul 11, 2015, at 9:48 PM, Christoph Pojer <notifications@github.com<mailto:notifications@github.com>> wrote:


Interesting. What I had on my mind was something like this:

node.replaceWith(
  parse`var $foo = $bar($args)`,
  {
    foo: j.identifier('javascript'),
    bar: j.identifier('is'),
    args: [j.identifier('awesome')],
  }
);

and it would print to var javascript = is(awesome);

What do you think about this approach? I'm a little worried your idea, while sound in theory, makes the API more complex as it introduces more ways of creating nodes.

fkling:
If you look at the source then you'll see that there is not much difference between `expression`, `statement` and `statements`. It's just a way to tell the tool which node to extract from the replacement string. I'd be happy about a simpler if there is one... but I don't necessarily see this as different ways of creating nodes.

src/collections/Node.js
```diff
@@ -143,8 +143,11 @@ var mutationMethods = {
 
     return this.forEach(function(path, i) {
       var newNode = replacement.call(path, path, i);
-      assert(Node.check(newNode), 'Replacement function returns a node');
-      path.replace(newNode);
+      if (Array.isArray(newNode)) {
+        path.replace.apply(path, newNode);
```
cpojer: `path.replace(...newNode);` with babel

package.json
```diff
@@ -32,13 +33,13 @@
...
   "devDependencies": {
+    "babel-jest": "^5.3.0",
     "es6-promise": "^2.0.1",
     "jest-cli": "^0.4.0",
-    "react-tools": "^0.13.1",
     "temp": "^0.8.1"
   },
   "jest": {
-    "scriptPreprocessor": "./scripts/test-preprocess.js",
+    "scriptPreprocessor": "./node_modules/babel-jest",
```
cpojer: so that fixes the jest issue?

cpojer:
I'm sorry, I was about to eat lunch and just wanted to leave that comment there without actually looking at the code :)

In my solution I liked that it fits more with pattern matching and actual templates but your solution looks more awesome and more straightforward. It wasn't obvious to me how it worked just from the summary but looking at the tests I love it, it is going to change how I write codemods.

I take it your vacation is officially boring now?

package.json
fkling: Not really, tests still fail with v0.12.

fkling:
I just had some me-time today ;) but I'm looking forward to work more on jscodeshift again.

I also was thinking about using placeholders first, but I then thought, why not use template literals now that we have them? This of course only works for replacements. I still want to have some kind of pattern matching for search, but don't know yet how to realize that. Thats why this is first step :)

fkling:
I think it would be great to come up with a list where we might want to treat an array of nodes in a special way. E.g., like in your example, if there is a placeholder is in the arguments position of a function call and the value is an array, treat each of them as an argument. That doesn't work currently.

package.json
cpojer: but they run in 0.10?

fkling: Yep

fkling:
Added some special treatment for places where we might want to explode an array of AST nodes, e.g.
`foo(${bar})` or `[${foo}, 42]`.

The tests run pretty slow, which probably means that parsing and traversing the code from the template literal. I suggest we just try it and see how it performs on real use cases.

But we can definitely keep @cpojer's suggestion in mind, which would allow us to compile the template *string* into a "template AST" *once* and then reuse it (much like lodash's `_.template` method).

```js
`foo = ${bar}`
```

would then simply be converted to something like

```js
var template = compile(template.join('${}')); // 'foo = ${}'
return template(nodes);
```

internally, where `compile` takes care of replacing the placeholders with variable names and converting the resulting source into an AST.

We just have to decided on a useful placeholder syntax than can easily be replaced. I thought something like `${}` for positional and `${foo}` for named placeholders, just like we would do with template literals, but that would conflict with template literals inside the template string:

```js
'var foo = 42; var ${x} = `${foo}`'
```

in the above example, `${foo}` is not a placeholder but part of the code.

I could also imagine `~foo~` instead.

src/__tests__/template-test.js
```diff
@@ -0,0 +1,180 @@
...
+
+    it('explodes arrays in call expressions', () => {
+      let input = `var foo = [a, b];`;
+      let expected = `var foo = bar(a, b, c);`;
+
+      expect(
+        jscodeshift(input)
+          .find('ArrayExpression')
+          .replaceWith(
+            ({node}) => expression`bar(${node.elements}, c)`
```
cpojer: hm, I think we might need a wrapper function for this. How do I differentiate between `bar([a, b], c)` and `bar(a, b, c)`?

Something like `${spread(node.elements)}` maybe?

cpojer:
Yeah, I think my solution would be more performant. What do you estimate the difference will be in real codemod? If it is 2x it is not an issue, if it is 10x it would suck on giant codebases.

I think we could turn your solution into mine internally since you are already using templates/random variables. You let babel transform the template once and then every time when it is being invoked you traverse the AST nodes and replace the temp variables with the actual values. It is just a little more work inside of jscodeshift instead of inside of babel.

If you think this is fine for now, feel free to merge!

src/__tests__/template-test.js
fkling: If you want to insert a literal array, you would simply build one:

    `bar(${jscodeshift.arrayExpression([...])})`

fkling:
> What do you estimate the difference will be in real codemod?

No idea, we'll see. And yea, that's exactly what I meant. Will merge once I'm back (and looked more into the test stuff).

hzoo:
@fkling Should I just close #15 since you also added babel to this PR - or redo my PR on top of this one?

fkling:
@hzoo: I just use Babel to convert jscodeshift itself, not for parsing the source. So, if you could redo yours on top of master after I merged this one, that would be great

hzoo:
Oh ok - I did it to your PR instead oops

fkling:
I guess either way would have worked :)

PR #17, opened 2015-07-14
Add support for arrays to replaceWith, insertBefore and insertAfter.

cpojer:
I've been working around this in my codemods so far, this makes it so that you can pass an array to these methods (or return an array from the replacer method). I also removed `functionify` and inlined it.

I also added tests for insertAfter.

### DIFF
--- a/src/collections/Node.js
+++ b/src/collections/Node.js
@@ -20,17 +20,6 @@ var recast = require('recast');
 var Node = recast.types.namedTypes.Node;
 var types = recast.types.namedTypes;
 
-
-function methodNameFromType(type) {
-  return 'visit' + type;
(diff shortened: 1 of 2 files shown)


fkling:
lgtm... seems like call spread is already supported?

cpojer:
Jest has trouble with the jscodeshift-test in `bin`. Not sure what is going on, do you have an idea? It also happens locally. And yeah, call spread is supported in babel.

fkling:
Will have to look into it when I'm back but looks like the jest issue we still haven't fixed.

But you are not using Babel ;)

fkling:
Updated for latest master.

PR #18, opened 2015-07-14
[RFC] Remove node

cpojer:
Removing a node is hard. This is an initial implementation of a function `remove` that can remove a node if it is part of any AST field that is an array. I probably missed some but this should be a good start.

I'm not entirely sure what I should do in the case that the parent becomes empty. In some cases you'd want the parent to remove itself (empty block in a function?) but sometimes you probably don't (removing all params from a function, remove all items from an object expression). I think this is best left to the developer, although I'm wondering if we can provide hints. Should the `remove` method return something like `All expressions from a "SequenceExpression" were removed. You can now consider removing the SequenceExpression`?

### DIFF
--- a/src/collections/Node.js
+++ b/src/collections/Node.js
@@ -165,6 +165,10 @@ var mutationMethods = {
         (typeof insert === 'function') ? insert.call(path, path, i) : insert;
       path.insertAfter(...toArray(newNodes));
     });
+  },
+
+  remove: function() {
+    this.forEach(path => path.prune());
   }
 
 };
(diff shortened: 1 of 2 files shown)


fkling:
What's wrong with `prune`? https://github.com/benjamn/ast-types/blob/master/lib/node-path.js#L60

fkling:
Also this includes changes from the other PR.

cpojer:
Wait, are you saying prune does all of this, just smarter?

I included the other changes because otherwise I'd have a conflict in the Node-test.js, and figured we'll merge the other one.

PR #19, opened 2015-07-16
update esprima-fb to 15001.1.0-dev-harmony-fb

voideanvalue:
See fkling/esprima_ast_explorer#14

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "dependencies": {
     "async": "^0.9.0",
     "cli-color": "^0.3.2",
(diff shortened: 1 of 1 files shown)


voideanvalue:
Not sure why the tests pass on Node 0.10 but fail on 0.12. I can reproduce the passing and failure locally with those versions. But the [error message](https://travis-ci.org/facebook/jscodeshift/jobs/71191163#L211) isn't helpful at all.

hzoo:
@voideanvalue I think it's related to the jest issues mentioned in other PRs

PR #20, opened 2015-07-16
also use babel for recast esprima parser

hzoo:
Good thing it's that simple to just do another PR

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "optional": ["runtime"]
+  "optional": ["runtime"],
+  "stage": 0
 }
--- a/src/collections/__tests__/JSXElement-test.js
+++ b/src/collections/__tests__/JSXElement-test.js
@@ -12,7 +12,7 @@
 
 jest.autoMockOff();
 
-var esprima = require('esprima-fb');
+var babel = require('babel');
 
 describe('JSXCollection API', function() {
   var nodes;
@@ -40,7 +40,7 @@ describe('JSXCollection API', function() {
       '  </Child>',
       '  <Child id="2" foo="baz"/>',
       '</FooBar>'
-    ].join('\n'), {esprima: esprima}).program];
+    ].join('\n'), {esprima: babel}).program];
   });
 
   describe('Traversal', function() {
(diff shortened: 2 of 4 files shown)


hzoo:
Ok nice passes - although I forgot `fixes #14` in the commit message

fkling:
Could you try to run the tool on code containing experimental stuff, such as `async/await` or class properties? Not sure if we have to set some options to be able to parse these.

benjamn:
Why `babel` and not just `babel-core`?

fkling:
@benjamn: I already included `babel` in 0469b0c40ab38ffb1e4a96e2b9fcf5782d21e194.

hzoo:
Um I can add to the `.babelrc` file you made for async

```js
{
  "optional": ["runtime"],
  "stage": 1
}
```

but if you want class properties then it would be stage 0

cpojer:
I think for a codemod tool, stage 0 is reasonable. Even better however would be a way to specify the options in the transform file.

hzoo:
@cpojer Ok you mean pass the options from the file itself/cli to babel instead of always being stage 0?

fkling:
To avoid confusion: Babel is used here in three different situations:

- To transform the jscodeshift source itself (already implemented in an earlier)
- To *parse* the files that are supposed to be codemodded (this PR)
- To transform the codemod files (future)

`.babelrc` is currently primarily used for the first case. To not conflate use cases, it would make sense IMO to pass options explicitly to the Babel API here.

However, it doesn't look like we have to set any specific options to *parse* code with experimental features, e.g.

```js
babel.parse('async function foo(){}')
```

works fine but

```js
babel.transform('async function foo(){}')
```

does not, indicating that we only need to set options for experimental stuff when we actually have to transform.

So this PR seems good to go. I will probably release another patch version with the current changes and bump minor for this PR, since it potentially breaks existing codemods.

fkling:
Since it wasn't clear from my comment: `state: 0` is not necessary. I'll remove it.

fkling:
Merged with 3d68222b6e43965e84f8a553c133d46ea4aba0d4.

fkling:
OK, now merged with f35058c42090c1a2a5b4194dfec1432ef61d0503.

hzoo:
Oh ok nice :+1:

PR #23, opened 2015-07-24
Pre-process transform files with babel. Fixes #5

### LINKED ISSUE #5: Bake in ability to preprocess the transform script
The motivating use case here is the ability to transform the transform script with a tool like Babel (or other) prior to use.

Just say you're working on a large project where you don't control the build process and so you can't integrate your transform scripts with it.

Or alternatively, say you're wanting to run a transform script that comes from somewhere else (eg. another repo checked out somewhere else on the filesystem), and again you can't transparently pre-build the transform.

Current workaround is to make a simple wrapper that does the transform first.

Question is, is it outside the scope of jscodeshift to make it easy to plug-in some preprocessing step for the transform prior to using it? Possible questions:

- should `jscodeshift` accept a transform script via standard input? (in which case you could do something like `babel myTransform.js | jscodeshift ....`)
- should we expose a `--preprocess` option that allows the user to specify an executable to pre-process the transform before running it?
- should we expose another mechanism to inject a JS module that does the preprocessing?
- what would the implications of this be for preprocessing the other files (ie. the ones that the transform operates on?)

cpojer:
This uses babel's require hook. The downside is that every worker will transform every file, but workers are limited to number of CPUs so I think the overhead of doing this a few times isn't a big deal. Also added an option `babel` to disable this behavior.

Other solutions we've explored:
* Transform the script in the Runner and create a temp file. This breaks `require` calls in the transform script because the temp file would be relative
* Put the temp file into the same folder as the transform file. This means that `require` works but other files won't be transformed.

### DIFF
--- a/bin/jscodeshift.sh
+++ b/bin/jscodeshift.sh
@@ -49,6 +49,11 @@ var opts = require('nomnom')
       abbr: 'p',
       flag: true,
       help: 'Print output, useful for development'
+    },
+    babel: {
+      flag: true,
+      default: true,
+      help: 'Do not apply babel for transform files'
     }
   })
   .parse();
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -127,7 +127,7 @@ function run(transformFile, files, options) {
   file_chunks.forEach(function(files) {
     var child = child_process.fork(
       require.resolve('./Worker'),
-      [transformFile]
+      [transformFile, options.babel ? 'babel' : 'no-babel']
     );
     child.send({files: files, options: options});
     child.on('message', onMessage);
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -10,6 +10,10 @@
 
 'use strict';
 
+if (process.argv[3] === 'babel') {
+  require('babel/register')();
+}
+
 var async = require('async');
 var fs = require('fs');
 


src/Runner.js
```diff
@@ -127,7 +127,7 @@ function run(transformFile, files, options) {
   file_chunks.forEach(function(files) {
     var child = child_process.fork(
       require.resolve('./Worker'),
-      [transformFile]
+      [transformFile, options.babel ? 'babel' : 'no-babel']
```
cpojer: I could use `true` and `false` or `1` and `0` or any other string but I think this is the most explicit and easiest to understand.

bin/jscodeshift.sh
```diff
@@ -49,6 +49,11 @@ var opts = require('nomnom')
       abbr: 'p',
       flag: true,
       help: 'Print output, useful for development'
+    },
+    'babel': {
+      flag: true,
+      default: true,
```
fkling: How do I disable transform? Does `--babel=false` work?

cpojer: yes, or use `--no-babel`

fkling: Oh, this adds `--no-<whatever>` implicitly? Neat.

src/Worker.js
```diff
@@ -10,6 +10,10 @@
 
 'use strict';
 
+if (process.argv[3] == 'babel') {
```
fkling: `===` for consistency? :)

fkling:
This still needs an explanation in the README.

PR #24, opened 2015-07-24
Add ability to pass one or many directories to jscodeshift. Fixes #21

### LINKED ISSUE #21: Transform a directory of files
`jscodeshift html/js --extensions=js,jsx` should transform all files recursively in the passed in directory that match the extensions option.

cpojer:
This allows us to use jscodeshift without find/xargs which should really help with the cli ergonomics. I've been fighting with this all the time and got really annoyed, so glad I finally got around to change the Runner.

This makes use of `node-dir` and uses Promises. I converted the Runner code to ES6, pretty sure there are some subjective code style choices but I tried to make the style more "functional" and use const where appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ jscodeshift --help
 
 Usage: jscodeshift <path>... [options]
 
-path     Files to transform
+path     Files or directory to transform
 
 Options:
    -t FILE, --transform FILE   Path to the transform file  [./transform.js]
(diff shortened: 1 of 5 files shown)


fkling:
Can you simply increase the timeout in `bin/__tests__/jscodeshift-test.js` so that this test passes as well?

pieterv:
Would be nice to use `node-glob` instead of `node-dir`, so you can do nice matching like `my-project/**/MyPrefix*.js`. Any reason why you didn't use it? if not i can do a PR if find the need for it :)

cpojer:
I didn't think that was important. In my codemods I always do the filtering in code (like search for specific require statements) and I just run them on everything. If you would prefer node-glob, feel free to send a PR.

pieterv:
Ok thanks @cpojer, ill see if it comes up and PR if needed :)

fkling:
I think one of the issues is also that your shell will evaluate the path before it's passed to jscodeshift. However, some shells, with the correct setting, actually support `**` natively. 
I'm not opposed to it, but I'd also say we just see if there is an actual need for this.

PR #27, opened 2015-07-30
Add a newline to the end of a file by default.

cpojer:
Multiple people have asked me why in their diff the last line of every file was changed. I think this is a reasonable default and will allow me to get rid of the `+ '\n'` code in most of my transforms.


cpojer:
Changed it to automatically detect whether it needs to add a newline.

cpojer:
See #29

PR #29, opened 2015-07-31
Fix printing

fkling:
There have been multiple reports about the printed source missing trailing line breaks. The reason for the line break missing is that this information is contained in the `File` node, which is returned by
`recast.parse`.
However, because we build a node collection from the `Program` node instead of the `File` node, we are loosing this information.
Using the `File` node fixes the issue and doesn't seem to break any existing functionality.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,11 @@
+{
+  "parser": "babel-eslint",
+  "rules": {
+    "comma-dangle": 0,
+    "quotes": [2, "single", "avoid-escape"],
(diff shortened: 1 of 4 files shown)


cpojer:
:+1: 

Seems like Travis is unhappy with the timeout, as always?

I originally followed @voideanvalue's suggestion that this should not be in recast, but given that it already seems to handle this it makes a lot of sense.

fkling:
Not again... I will just set the timeout to a minute :P

PR #31, opened 2015-08-06
Fix for "cpus" command line parameter

mjomble:
There is no "cpus" variable at that point, which breaks the Math.min() expression and as a result, no workers are spawned if the -c option is used.

"availableCpus" seems to be the variable that was meant to be used and fixed the problem for me.


cpojer:
thanks!

fkling:
Published in v0.3.4

PR #38, opened 2015-08-12
Only apply filters on files from a directory, fixes #32

### LINKED ISSUE #32: Anly apply extension filters to recursive directory traversal, not directly passed files

cpojer:


src/Runner.js
```diff
@@ -90,11 +93,10 @@ function run(transformFile, paths, options) {
...
   }
 
-  getAllFiles(paths)
-    .then(files => files.filter(
-      name => !extensions || extensions.indexOf(path.extname(name)) != -1
-    ))
-    .then(files => {
+  getAllFiles(
+    paths,
+    name => !extensions || extensions.indexOf(path.extname(name)) != -1
```
fkling: Thinking about it, this is kind of weird IMO. `js` would also match `jsx` and `json`. I think this should be stricter.

cpojer: good point.

cpojer: wait, no, you are not right.

extensions is an array and the value passed to indexOf is the extension name of the file.

fkling: ooooh, ok then :) (I blame JavaScript)

PR #51, opened 2015-09-07
Remove check on registerMethods

aaronjensen:
This check prevents using registerMethods with iterative development with things like test:watch, astexplorer.net, etc.

Maybe there's a better way to handle this, but I'm not sure the check provides much value period and it has caused me plenty of headaches.

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -298,9 +298,6 @@ var CPt = Collection.prototype;
  */
 function registerMethods(methods, type) {
   for (var methodName in methods) {
-    if (CPt.hasOwnProperty(methodName)) {
-      throw Error(`A method with name "${methodName}" already exists.`);
-    }
     if (!type) {
       CPt[methodName] = methods[methodName];
     } else {


jamestalmage:
I think the check has value. #108 provides an alternative solution (users provide plugin callbacks, which will only be called once). It may not solve the `astexplorer.net` issue. If not, that feels like a pretty unique circumstance - so maybe only disable this check on that site.

aaronjensen:
@jamestalmage i'm not sure i understand what #108 has to do with this. An alternative solution to what? The problem here? It seems like another layer of unnecessary checks. What value does this check actually have? It strikes me as pure defensive coding with a very thin upside that makes things painful for anyone trying to use `registerMethods` in tests or astexplorer.net.

jamestalmage:
Using #108, you would register your extension like so:

```js
function plugin(jscodeshift) {
  jscodeshift.registerMethods({
    // method definitions here
  }); 
}

jscodeshift.use(plugin); // extensions are registered
jscodeshift.use(plugin); // has no effect
```

It prevents duplicate registration. This pattern is used by other extensible libraries (`chai` for instance) for exactly the same reason.

> It seems like another layer of unnecessary checks.

It adds no new checks. It provides a utility for you to avoid calling the same method registration code twice.

> It strikes me as pure defensive coding with a very thin upside 

Defensive checks have value. It's entirely conceivable that someone might mistakenly try to register two methods with the same name, but different behaviors. On large projects/teams you might not even be aware the other extension exists.

fkling:
I'm closing this since we don't have the issue in astexplorer anymore and we will likely move in the direction of #109.

Thanks for the discussion around this!

PR #61, opened 2015-10-21
Improvements to matchNode

yungsters:
Makes a couple improvements to `matchNodes`:

- Allow `path.find(j.Identifier, node => WHITELIST.hasOwnProperty(node.name))`.
- Fix fatal in `path.find(j.Identifier, {loc: null})`;

Also, added unit tests.

### DIFF
--- a/src/__tests__/matchNode-test.js
+++ b/src/__tests__/matchNode-test.js
@@ -1,0 +1,109 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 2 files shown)


cpojer:
thank you! Looks good. I'll let Felix merge and tag the new release, because he has the powers!

yungsters:
Updated this to support functions at any level:

```
matchNode(..., {
  type: 'MemberExpression',
  object: {
    type: 'Identifier',
    name: name => name === 'foo' || name === 'bar'
  },
});
```

Also, added unit tests.

fkling:
> Also, added unit tests.

:+1:

PR #62, opened 2015-10-21
Upgrade to Jest v0.5

yungsters:


cpojer:
You need to update .travis.yml, otherwise this won't work: https://github.com/facebook/jscodeshift/blob/master/.travis.yml#L4

fkling:
You also need to update `.travis.yml`.

yungsters:
Updated to node v4.

package.json
```diff
@@ -37,7 +37,7 @@
   "devDependencies": {
     "babel": "^5.6.14",
     "babel-jest": "^5.3.0",
-    "jest-cli": "^0.4.0",
+    "jest-cli": "^0.5.0",
```
cpojer: I don't quite understand how npm's versioning works but jest only started working with node 4 at around 0.5.6 or something. It might make sense to specify `0.5.10` as the minimum requirement.

yungsters: Done.

PR #69, opened 2015-11-07
runner promise to be fullfilled with file counters

caridy:
```js
Runner.run(
          transformer,
          files,
          opts
).then(function (result) {
          // this promise was always throwing, and result was not reacheable
})
```

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -154,7 +154,7 @@ function run(transformFile, paths, options) {
           endTime[0],
           (endTime[1]/1000000).toFixed(0)
         );
-        resolve();
+        return fileCounters;
       })
     );
 }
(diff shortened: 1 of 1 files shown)


src/Runner.js
```diff
@@ -154,7 +154,7 @@ function run(transformFile, paths, options) {
           endTime[0],
           (endTime[1]/1000000).toFixed(0)
         );
-        resolve();
```
caridy: `resolve` is not colleable, it is just the array with the results of `pendingWorkers` promises.

resolving to `fileCounters` seems more useful when runner is called programmatically instead of via CLI.

fkling: Thanks for that! Could you please also remove `resolve` from the function signature, to avoid confusion?

caridy: done

fkling:
Added with 3527f7c8017a0d197ac5ab42bd84021c42561d43 (I merged the commits) and published with v0.3.9 .

caridy:
awesome, thanks!

PR #71, opened 2015-11-21
Add '--run-in-band' option to run in the current process

zertosh:
Like jest's `--run-in-band` so you can use `--debug/-brk` without hating life. If the `if (module.parent) {` block is too gross, then I'm open to suggestions. cc: @fkling @cpojer


fkling:
Sorry for the late reply, that seems reasonably to me. Will take a closer look soon.

zertosh:
Hey @fkling, have you had a chance to take a look?

fkling:
I'll get to it over the weekend! You are right, it's not "pretty" but I don't think there will be a much better way anyway.

PR #77, opened 2015-12-14
binary file name changed

udnisap:


fkling:
yeah, that makes more sense, however I think this could potentially be a breaking change, so I will leave this till the next breaking version.

udnisap:
@fkling I think this is not a breaking change as the updated version will use the updated package.json which has the correct link

Daniel15:
Is this still an issue?

ElonVolo:
There don't seem to be any other complaints about this in 7 years or so since the PR was submitted. The context is so old at this point it's probably just worth closing.

PR #83, opened 2016-01-03
Fix `Collection.closest`

juliankrispel:
At the moment `Collection.closest` doesn't actually work with filters. If `type.check(parent.value)` in the while [loop on line 80](https://github.com/facebook/jscodeshift/blob/master/src/collections/Node.js#L80) returns `true` it will stop the while loop, skipping the last expression, which tests if the filter condition applies.

This pr fixes this and refactors the while loop into a recursive function which is imo more readable.

I also added some specs to prove that closest is working correctly.

### DIFF
--- a/src/collections/Node.js
+++ b/src/collections/Node.js
@@ -19,30 +19,6 @@ var recast = require('recast');
 var Node = recast.types.namedTypes.Node;
 var types = recast.types.namedTypes;
 
-const findParent = (path, type, filter) => {
-  var parent = path.parent;
-  if (!parent) {
-    // Return null if we've reached the upper bound of
-    // the AST
-    return null;
-  } else if (
-    // return the node if we haven't specified a filter
-    // and the type check returns true.
-    parent && 
-    type.check(parent.value) && !filter) {
-    return parent;
-  } else if(
-    // return the node if both type check and filter
-    // return true
-    parent && type.check(parent.value) &&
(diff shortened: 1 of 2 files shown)


fkling:
Thanks for noticing this! However, I'm not in favor of a recursive solution. It's unlikely that someone will run in to call stack issues, but still. I'm more comfortable doing this when we have TCO :)

It seems this could be easily fixed by changing the `while` condition to

```javascript
while (
  parent &&
  !(
    type.check(parent.value) &&
    (!filter || matchNode(parent.value, filter))
  )
) {
```

(or something equivalent)

juliankrispel:
@fkling my pleasure, thanks for this awesome project! I'll change it to your liking, it's your project after all, all I want to do is contribute :)

Just as a side note: I'm not strongly opinionated about this, but it certainly helped to debug the thing, also three if conditions although more verbose seemed easier to reason about than one big condition, hence I thought it was reasonable.

juliankrispel:
@fkling changed it back to a while loop, the added tests cover the issue I noticed so it should be good to go :)

fkling:
I squashed all the commits. This is now merged with faba4970ef37231be0bf4a769a21c0d6dd1cbd56. Thank you again! :)

juliankrispel:
@fkling thanks :)

juliankrispel:
@fkling thanks :), here's a kitten

image

Issue #84, opened 2016-01-15
jscodeshift Using .babelrc project file incorrectly in Babel 6 Environments


kjanoudi:
Scenario: 
* Project using Babel 6, and .babelrc file for project configuration
* Babel 6 installed globally
* jscodeshift is run within the project directory at the same level as the .babelrc file

Result:

```
/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/transformation/file/options/option-manager.js:126
      if (!option) this.log.error("Unknown option: " + alias + "." + key, ReferenceError);
                           ^

TypeError: Cannot read property 'error' of undefined
    at OptionManager.mergeOptions (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/transformation/file/options/option-manager.js:126:28)
    at OptionManager.addConfig (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/transformation/file/options/option-manager.js:107:10)
    at OptionManager.findConfigs (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/transformation/file/options/option-manager.js:168:35)
    at OptionManager.init (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/transformation/file/options/option-manager.js:229:12)
    at compile (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/api/register/node.js:117:22)
    at normalLoader (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/api/register/node.js:199:14)
    at Object.require.extensions.(anonymous function) [as .js] (/usr/local/lib/node_modules/jscodeshift/node_modules/babel-core/lib/api/register/node.js:216:7)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
```

Temporary fix:
* Run jscodeshift from outside the project directory, with no .babelrc file in the directory you are running jscodeshift within.

kjanoudi:
This looks like it has to do with https://github.com/facebook/jscodeshift/blob/master/src/Worker.js#L41 

```javascript
function setup(tr, babel) {
  if (babel === 'babel') {
    require('babel-core/register')();
  }
  transform = require(tr);
}
```

With babel 6+, the recommended approach is to use ``require('babel-register')({options})``. 

[...]

PR #85, opened 2016-01-23
Disable external Babel configuration

sviridov:
Hi!

This PR prevents jscodeshift's Babel from picking up external Babel configuration.
See [this](https://github.com/facebook/jscodeshift/issues/84) issue for more information.

### DIFF
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -40,7 +40,8 @@ if (module.parent) {
 
 function setup(tr, babel) {
   if (babel === 'babel') {
-    require('babel-core/register')();
+    // FIXME: use { babelrc: false } after migration to Babel 6
+    require('babel-core/register')({ breakConfig: true });
   }
   transform = require(tr);
 }


fkling:
YOLO: https://phabricator.babeljs.io/T1633 . It seems that `babelrc: false` should have the same effect. Would you mind trying that instead?

sviridov:
Yeah, I tried to use it, but it's not working. The purpose of `babelrc` has been changed in Babel 6.

[Babel 5](https://github.com/babel/babel.github.io/blob/862b43db93e48762671267034a50c30c00e433e2/docs/usage/options.md):

```
babelrc (default value = null) Specify a custom file path to load a .babelrc config from.
```

[Babel 6](https://babeljs.io/docs/usage/options/):

```
babelrc (default value = true) Specify whether or not to use .babelrc and .babelignore files.
```

fkling:
Uh :-/ It looked like it should work because of [this check](https://github.com/babel/babel/blob/5.x/packages/babel/src/transformation/file/options/option-manager.js#L180-L183):

```javascript
    // resolve all .babelrc files
    if (opts.babelrc !== false) {
      this.findConfigs(opts.filename);
    }
```

But OK, `breakConfig` is basically doing the same thing a few lines above.

sviridov:
Well, `opts.babelrc` is `undefined`... I did very simple thing:

```javascript
function setup(tr, babel) {
  if (babel === 'babel') {
    require('babel-core/register')({ babelrc: false });
  }
  transform = require(tr);
}
```

```javascript
console.log('opts:', opts);
// resolve all .babelrc files
if (opts.babelrc !== false) {
  this.findConfigs(opts.filename);
}
```

Output: `opts: { filename: '<my-transformation>.js' }`.
And they taking `breakConfig` not from `opts` but from `this.options`...

fkling:
Ah I see. The options object passed to the function is not passed passed to `init`, but added [separately](https://github.com/babel/babel/blob/5.x/packages/babel/src/api/register/node.js#L77):

```javascript
  var optsManager = new OptionManager;
  optsManager.mergeOptions(transformOpts);
  opts = optsManager.init(opts);
```

that's why `opts.babelrc` doesn't work but `this.options.breakConfig` does.

fkling:
Released with v0.3.13.

sviridov:
Awesome, thanks!

PR #90, opened 2016-02-27
Stream

Haurya:
Replaced console.logs with readable streams. Issue 80.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -93,8 +93,7 @@ function run(transformFile, paths, options) {
 
   if (!fs.existsSync(transformFile)) {
     rs.push(
-      clc.whiteBright.bgRed('ERROR') + ' Transform file %s does not exist\n',
-      transformFile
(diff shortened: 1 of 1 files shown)


fkling:
Thanks for taking is on! I believe your solution is a bit more complicated that it has to be. I don't see why you'd provide a writeable stream and pipe it to stout. You could just write to stdout directly!

However, overall I had a different solution in mind: Instead of having the runner produce any output, it would simply emit events with information about its status. Then whatever calls `run` could decide how to react to these events:

- The CLI will write a readable version of the statuses to stdout
- Custom reporters could collect the information print / return a JavaScript object (one of many possibilities)
- Unit tests could listen to the events for assertions.

However, that would require a bit more refactoring I believe, not sure you want to do that. If yes, I'd be very happy :)
If not, moving away from `console.log` is still a good idea, but I would rather suggest a solution closer to your previous PR, where you directly wrote to stdout.

Either way, thank you very much for your help!

PR #91, opened 2016-03-02
swapped all console.logs for stdout.write

Haurya:
swapped all console.logs for stdout.write


fkling:
Committed with 3968c2e051b2d455e0293f00d018c01aa9f39a52. Thanks!

PR #92, opened 2016-03-06
fix symlink problems

avikchaudhuri:
Re-implement the function `dir.files` that walks directories to collect files.

The main change is that when stat problems are encountered (e.g., due to symlink issues), we log an error but continue on, instead of refusing to any work.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -64,19 +64,23 @@ function dirFiles (dir, callback, acc) {
   }
 
   fs.readdir(dir, (err, files) => {
+    // if dir does not exist or is not a directory, bail
+    // (this should not happen as long as calls do the necessary checks)
     if (err) throw err;
+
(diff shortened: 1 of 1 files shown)


zertosh:
I've had similar problems before, but I ended up going with `find ... | xargs jscodeshift` since that gives me way more control (e.g. exclude `node_modules`). Just thinking out loud, I wonder if `jscodeshift` should be in the business of directory walking at all?

cpojer:
@zertosh it was added mainly as a convenience. I think jscodeshift should provide a reasonable default.

zertosh:
Fair enough!

fkling:
One issue with `find` and `xargs` is that if you have too many files, `xargs` will spawn multiple `jscodeshift` processes. But I guess we can fix that too by letting `jscodeshift` read from `stdin`.

Issue #87, opened 2016-02-05
Per File Timeout


ForbesLindesay:
It would be really useful to have a per-file timeout option.  Currently I'm transforming a large number of files, and some individual files are causing my transform to hang (seemingly indefinitely).  If I could set a timeout on a per-file basis then I could just continue on and transform all other files.

This could work by killing any process which does not produce any output for a sufficiently long time, then putting whatever jobs it had pending into a new process.

fkling:
That seems reasonable to me.

lirantal:
Can this be closed due to related merged PRs?

PR #93, opened 2016-03-08
worker optimization

avikchaudhuri:
Change the worker management to be dynamic rather than static. Instead of dividing all files among the available number of processes, we now send chunks of work to each free process, and when a process finishes its chunk of work it gets another chunk. This leads to overall much better perf for a large number of files with non-uniform processing times.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -19,6 +19,7 @@ const fs = require('fs');
 const path = require('path');
 
 const availableCpus = require('os').cpus().length - 1;
+const CHUNK_SIZE = 50;
 
 const log = {
   ok(msg, verbose) {
@@ -80,7 +81,6 @@ function run(transformFile, paths, options) {
   const cpus = options.cpus ? Math.min(availableCpus, options.cpus) : availableCpus;
   const extensions =
     options.extensions && options.extensions.split(',').map(ext => '.' + ext);
-  const fileChunks = [];
   const fileCounters = {error: 0, ok: 0, nochange: 0, skip: 0};
   const statsCounter = {};
   const startTime = process.hrtime();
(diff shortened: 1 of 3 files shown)


src/Runner.js
```diff
@@ -19,6 +19,7 @@ const fs = require('fs');
 const path = require('path');
 
 const availableCpus = require('os').cpus().length - 1;
+const CHUNK_SIZE = 50;
```
avikchaudhuri: Should I make this a config option?

src/Worker.js
```diff
@@ -74,6 +78,9 @@ function trimStackTrace(trace) {
 function run(data) {
   var files = data.files;
   var options = data.options;
+  if (!files.length) {
+    finish();
+  } else
```
avikchaudhuri: Should format whatever comes next properly. Will fix.

src/Runner.js
fkling: I'm fine with this being a constant for the time being.

fkling:
Sweet, thank you! Will merge your PRs tomorrow.... maybe this also makes it a bit easier to solve #87.

avikchaudhuri:
Not sure why the build failed, tests pass locally. Any way to kick off another build?

cpojer:
btw. I'd recommend switching to node-worker-farm: https://github.com/rvagg/node-worker-farm

fkling:
I just restarted it. Sometimes testing the Babel option fails, no idea why.

avikchaudhuri:
Thanks! @cpojer Is it OK if we merge this PR now and we take on the work to use node-worker-farm later? I'm willing to investigate that option (looks really promising) but meanwhile I need some moar perf! :)

fkling:
Released as v0.3.14. Thank you!

PR #96, opened 2016-03-14
Fix processes count and chunk size for run-in-band mode

zertosh:
The worker was being called more than once when in run-in-band mode.


fkling:
Released with v0.3.15.

PR #101, opened 2016-03-18
Fix encoding errors in logging messages

cjlarose:
The [`write` method on Stream][write] takes the second argument to be an encoding. This is certainly not what's needed in the log messages.

If you run `jscodeshift -v 2` on a file, you can get `TypeError` exceptions.

[write]: https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback


fkling:
Whoops. That was probably from my unsuccessful merge :-/ Thanks!

PR #103, opened 2016-03-26
cli-color -> colors

pieterv:
Swap out `cli-color` dep for `colors` since it has no dependencies.

I'm having a lot of problems with `cli-color` deps `es5-ext` and `es6-iterator` since they have folders named `#` which gets dropped by mercurial by default. Additionally these deps have non utf8 files which are causing problems. It seems kinda overkill when all we want are some color codes so lets just use the simple `colors` module.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -13,7 +13,7 @@
 require('es6-promise').polyfill();
 
 const child_process = require('child_process');
-const clc = require('cli-color');
+const colors = require('colors/safe');
 const fs = require('fs');
 const path = require('path');
 const http = require('http');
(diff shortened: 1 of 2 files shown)


src/Runner.js
```diff
@@ -99,7 +99,7 @@ function getAllFiles(paths, filter) {
     paths.map(file => new Promise((resolve, reject) => {
       fs.lstat(file, (err, stat) => {
         if (err) {
-          process.stdout.error('Skipping path ' + file + ' which does not exist. \n');
```
pieterv: This was giving me errors when testing, as far as i can tell node doesn't have a `stdout.error`. So i just updated to `stderr.write`.

fkling: That's the result of a bad merge from my side. I didn't replace those instances properly :-/ Thanks for fixing that!

fkling:
I should pay closer attention to the dependencies of dependencies...

cpojer:
It seems like `chalk` is kind of the spiritual successor to `color`: https://www.npmjs.com/package/chalk

pieterv:
@cpojer  `chalk` looks nice although i didn't see anything that we really need that `colors` can't do, plus `chalk` has 9 deps (they did look pretty small though). Also i saw this site when i was looking into this stuff, its pretty cool: http://npm.anvaka.com/#/view/2d/cli-color.

Thanks for the quick review @fkling!

PR #104, opened 2016-04-05
Add simple utility for unit testing jscodeshift transform

Daniel15:
Adds a simple utility to allow unit testing of jscodeshift transforms in a standard way, without having to write a bunch of boilerplate code. This is loosely based off [the technique used in react-codemod](https://github.com/reactjs/react-codemod/blob/master/jest/env.js#L23), with some changes. The main change is that where possible it doesn't depend on any global environment/configuration:

 - It's in an actual module rather than a global function called "test". It took me *so long* to work out where the `test` function was coming from in the `js-codemod` repo :stuck_out_tongue: 
 - Assumes the transform is in the parent directory of the test (eg. transform is at `foo/bar.js`, test is at `foo/__tests__/bar-test.js`)
 - Does not use a `.js` suffix for the test fixtures (input and expected output). This is mainly so they don't accidentally get picked up as unit tests by Jest, or as code that should be bundled with the app.

Also included an example of how to use it, using a simple transform (the default example from AST Explorer).

### DIFF
--- a/src/testUtils.js
+++ b/src/testUtils.js
@@ -1,0 +1,78 @@
+/**
+ *  Copyright (c) 2016-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/* global expect, define, it */
+
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Utility function to run a jscodeshift script within a unit test. This makes
+ * several assumptions about the environment:
+ *
+ * - `dirName` contains the name of the directory the test is located in. This
+ *   should normally be passed via __dirname.
+ * - The test should be located in a subdirectory next to the transform itself.
+ *   Commonly tests are located in a directory called __tests__.
+ * - `transformName` contains the filename of the transform being tested,
+ *   excluding the .js extension.
+ * - `testFilePrefix` optionally contains the name of the file with the test
+ *   data. If not specified, it defaults to the same value as `transformName`.
+ *   This will be suffixed with ".input.js" for the input file and ".output.js"
+ *   for the expected output. For example, if set to "foo", we will read the
+ *   "foo.input.js" file, pass this to the transform, and expect its output to
+ *   be equal to the contents of "foo.output.js".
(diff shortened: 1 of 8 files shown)


src/testUtils.js
```diff
@@ -0,0 +1,74 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/* global expect, define, it */
+
+import fs from 'fs';
+import path from 'path';
+
+import jscodeshift from './core';
```
cpojer: this should be a lazy require inside of `runTest`. Jest resets the module registry after each test and we had issues previously where two copies of ast-types was present.

cc @pieterv

src/testUtils.js
```diff
@@ -0,0 +1,74 @@
...
+    options || {}
+  );
+  expect((output || '').trim()).toEqual(expectedOutput.trim());
+}
+
+/**
+ * Handles some boilerplate around defining a simple jest/Jasmine test for a
+ * jscodeshift transform.
+ */
+export function defineTest(dirName, transformName, options, testFilePrefix) {
```
cpojer: why not just call it `test`? :)

src/testUtils.js
Daniel15: Oooh, good point. Will fix.

src/testUtils.js
Daniel15: I wanted to differentiate between defining a test (calling `define` and `it`) and running a test (actually running the transform and checking the result), thus the explicit function naming.

src/testUtils.js
Daniel15: Updated.

fkling:
Awesome! Would you mind mentioning this in the readme somewhere?

Daniel15:
Sure :) I was thinking about it a bit, and I think I'll move the test fixtures to a separate directory to keep them separate from the unit test files themselves.

Daniel15:
Updated to add a section to the readme.

fkling:
Thanks, this is great!

Daniel15:
@fkling  - It would be useful to mention this in the [release notes for v0.3.20](https://github.com/facebook/jscodeshift/releases/tag/v0.3.20), mainly so people know that this is only included from that version onwards.

fkling:
Whoops, yes of course! Sorry for forgetting to add this!

Daniel15:
No worries :+1:

PR #107, opened 2016-04-16
Can ignore paths before jscodeshift runs

### LINKED ISSUE #53: Allow excluding subdirectories like node_modules
I'd like to run a mod on a large directory of directories, each of which can have a 'node_modules' directory I'd like to ignore. It'd be great to have an option in jscodeshift to exclude subdirectory patterns if you pass it a directory to transform.

chrisdarroch:
Fixes #53.

Usage:

```
# will ignore paths matching the glob
jscodeshift -t transform.js --ignore-pattern="*-test.js" ~/path/to/repo

# will ignore all files gitignore does
jscodeshift -t transform.js --ignore-config ~/path/to/repo/.gitignore ~/path/to/repo
```

The implementation could probably be made more robust by pulling in another package like `gitignore-to-glob`, but I tried to keep the dependency tree pretty much as-is.

### DIFF
--- a/bin/__tests__/jscodeshift-test.js
+++ b/bin/__tests__/jscodeshift-test.js
@@ -180,6 +180,30 @@ describe('jscodeshift CLI', () => {
     );
   });
 
+  describe('ignoring', () => {
+    var transform = createTransformWith(
+      'return "transform" + fileInfo.source;'
+    );
+    var sources = [];
+
+    beforeEach(() => {
+      sources = [];
+      sources.push(createTempFileWith('a', 'a.js'));
+      sources.push(createTempFileWith('a', 'a-test.js'));
+      // sources.push(createTempFileWith('b', 'src/lib/b.js'));
+    });
+
+    pit('does not transform files matched with --ignore-pattern', () => {
+      var pattern = '*-test.js';
+      return run(['-t', transform, '--ignore-pattern', pattern, ...sources]).then(
+        ([stdout, stderr]) => {
+          expect(fs.readFileSync(sources[0]).toString()).toBe('transforma');
+          expect(fs.readFileSync(sources[1]).toString()).toBe('a');
+        }
+      );
+    });
+  });
+
   describe('output', () => {
     pit('shows workers info and stats at the end by default', () => {
       var source = createTempFileWith('a');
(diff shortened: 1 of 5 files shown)


chrisdarroch:
@fkling what are your thoughts on the new flags? On the implementation? Is there anything you'd like to see changed or improved before merging?

bin/__tests__/jscodeshift-test.js
```diff
@@ -44,11 +45,23 @@ function run(args, stdin, cwd) {
...
+    return filePath;
+  }
+
+  function renameFileTo(oldPath, newFilename) {
+    var projectPath = path.dirname(oldPath);
+    var newPath = path.join(projectPath, newFilename);
+    mkdirp.sync(path.dirname(newPath));
+    fs.renameSync(oldPath, newPath);
+    return newPath;
   }
```
chrisdarroch: Renaming files in this way prevents the `temp` module from cleaning up the files after the test suite is finished running. Then again, we weren't calling `temp.track()` or `temp.cleanupSync()` in there. We might need to do that. 

Creating temp files with known names with the `temp` module isn't straightforward, since it only allows specifying prefixes and suffixes for file and directory names as opposed to the full and complete thing. We may need to manually clean up any files that run through this method ourselves.

fkling:
Sorry for the late reply, I have been on vacation.

This looks pretty good to me! I'm not a *huge* fan of `--ignore-pattern`, I could imagine `--ignore` would suffice, but OTOH, I like that it's explicit. Maybe we can just add a `-i` as well for ease of use?

Either way, I think it would make sense to declare `--ignore-pattern` as `list: true`, so that it can be specified multiple times.

chrisdarroch:
Yeah, I wasn't sure what to do with the difference between `ignore-pattern` and `ignore-config`. I originally thought to have just one parameter, but quickly realised I'd have no way to distinguish between a case of "I want to load config from this file" versus "please ignore just this file".

Changing `ignore-pattern` to a list seems sensible. I'll add a test for it and update the PR :)

fkling:
Committed as 5d58f4170394c78784bc899fa5f1207ca227cca0 .

danielo515:
So, the way to ignore multiple files is by passing multiple ignore patterns?

PR #108, opened 2016-04-20
Provide plugin registration method that avoids duplicate registrations.

jamestalmage:
Provides a `use` method on core that can be used to register extensions.
It avoids calling the same extension multiple times which is useful during testing.


fkling:
This seems reasonable to me, thank you!

PR #110, opened 2016-04-22
Allow non-conflicting method registrations. Fixes #109.

### LINKED ISSUE #109: Allow duplicate method registrations for non-conflicting types.
It's annoying that `renameTo` is an unavailable for non `VariableDeclarator`s. It seems to me that reuse should be allowed as long as the additional registration specifies a type that is not a super-type nor a sub-type of `VariableDeclarator`.

jamestalmage:
This allows you to register multiple typed methods with the same methodName, as long as there is no type overlap.

I've only added one additional unit test, though it might be wise to add a few more for the `hasConflictingRegistration` method.

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -311,7 +311,11 @@ function registerMethods(methods, type) {
       if (!CPt.hasOwnProperty(methodName)) {
         installTypedMethod(methodName);
       }
-      CPt[methodName].typedRegistrations[type] = methods[methodName];
+      var registrations = CPt[methodName].typedRegistrations;
+      registrations[type] = methods[methodName];
+      astTypes.getSupertypeNames(type).forEach(function (name) {
+        registrations[name] = false;
+      });
     }
   }
 } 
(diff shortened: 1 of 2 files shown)


jamestalmage:
Added a few more tests. My initial implementation did indeed have a bug (it only prevented registering an existing registration for a super-type existed, existing sub-type registrations did not prevent installation).

fkling:
This is awesome thank you!

src/Collection.js
```diff
@@ -298,26 +298,80 @@ var CPt = Collection.prototype;
  */
 function registerMethods(methods, type) {
   for (var methodName in methods) {
-    if (CPt.hasOwnProperty(methodName)) {
-      throw Error(`A method with name "${methodName}" already exists.`);
+    if (!methods.hasOwnProperty(methodName)) {
+      return;
+    }
+    if (hasConflictingRegistration(methodName, type)) {
+      throw Error(`There is a conflicting registration for method with name "${methodName}".`);
```
fkling: What do you think, would it be valuable to provide more information and mention the types which already have a registered method with that name?

jamestalmage: Updated. The error message is now much more descriptive.

jamestalmage:
@fkling - bump!

I'd like to start relying on this and #108 in some of my projects. Anything else I need to do to get this merged? How often do you release to `npm`?

fkling:
> How often do you release to npm?

Depends on my time and how big the changes are ;) Looks all good to me, I will cut a new release now!

fkling:
Merged in 81ba196707f2034e0f4f26d1daab3a781a13636a .

jamestalmage:
Awesome!

fkling:
Released v0.3.20 !

PR #113, opened 2016-06-10
Added tests for build and test npm scripts in travis ci

amilajack:


amilajack:
Finished signing.

facebook-github-bot-2:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

fkling:
I'm actually about to remove the build step... but I guess the addons and env options are still useful? I don't really know much about Travis CI config. What's the problem that this solves (honest question, because Travis CI seems to pass and test also pass locally)?

amilajack:
It passes because it doesn't have any tests. And travis ci is useful for ensuring coverage from external commits, PR's, etc. IMO, Travis CI's are worth having.

fkling:
> It passes because it doesn't have any tests.

It does have tests:

- https://github.com/facebook/jscodeshift/tree/0c7c25874e3b5b0d88db1a53e6064d3c6dcb21ae/src/__tests__
- https://github.com/facebook/jscodeshift/tree/0c7c25874e3b5b0d88db1a53e6064d3c6dcb21ae/bin/__tests__

> And travis ci is useful for ensuring coverage from external commits, PR's, etc. IMO, Travis CI's are worth having.

I never meant to imply anything else. That's why we added it to this repo.

---

Maybe my comment wasn't clear: Why exactly did you make the changes you made? Which problem are they solving? I'm just trying to understand why these configuration options are necessary.

amilajack:
My bad. I didn't see that the tests were running.

PR #114, opened 2016-06-11
Allow return many files from transform

typeetfunc:
Type: feature proposal
Description: allow return array from `transform` function. Array must contains objects with `path` and `source` props.
Example:
`return [{path: '/path/to/file1', source: 'someContent1'}, {path: '/path/to/file2', source: 'someContent2'}]`
Stage: draft
Also I want allow pass one object in array without prop `path` or just string i.e:
`return [{source: 'someContent1'}, {path: '/path/to/file2', source: 'someContent2'}]`
or
`return ['someContent1', {path: '/path/to/file2', source: 'someContent2'}]`
In this case `path` be equal to path source file.
Also need to add some checks for array of files.
Motivation: many kinds of refactoring include split big one file to many files. With ability to return many files will be able to make codemods such as:
- `codemod-split-utils-file` - I will be split file with many exported functions to many files according the rule "one file - one function". Source file will used as entry point and contatins only `export from` statements.
- `codemod-split-components-to-files` - It is the same as the previous but for React component
- `codemod-export-complex-flow-type` - It move big and complex type annotations in separate file
  Also I want add to [codemod-proptypes-to-flow](https://github.com/billyvg/codemod-proptypes-to-flow) option for place generated annotation in separate file(something like `Component.js.flow` for `Component.jsx`).

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ Options:
    -p, --print                 Print output, useful for development
    --babel                     Apply Babel to transform files  [true]
    --extensions                File extensions the transform file should be applied to  [js]
+   --ignore-pattern            Ignore files that match a provided glob expression
+   --ignore-config FILE        Ignore files if they match patterns sourced from a configuration file (e.g., a .gitignore)
+   --run-in-band               Run serially in the current process  [false]
+   -s, --silent                No output  [false]
 ```
 
 This passes the source of all passed through the transform module specified
--- a/bin/__tests__/jscodeshift-test.js
+++ b/bin/__tests__/jscodeshift-test.js
@@ -182,6 +182,26 @@ describe('jscodeshift CLI', () => {
     );
   });
 
+  pit('transform can return list of files', () => {
+    var source = createTempFileWith('a_b');
+    var secondFile = temp.path({suffix: '.js'});
+
+    var transform = createTransformWith(
+      `return fileInfo.source.split('_')
+        .map((source, i) => i === 0 ?
+          {source, path: fileInfo.path} :
+          {source, path: '${secondFile}'})`
+    );
+
+    return run(['-t', transform, source]).then(
+      ([stdout, stderr]) => {
+        expect(fs.readFileSync(source).toString()).toBe('a');
+        expect(fs.readFileSync(secondFile).toString()).toBe('b');
+      }
+    );
+  });
+
+
   describe('ignoring', () => {
     var transform = createTransformWith(
       'return "transform" + fileInfo.source;'
--- a/bin/__tests__/jscodeshift-test.js
+++ b/bin/__tests__/jscodeshift-test.js
@@ -182,14 +182,14 @@ describe('jscodeshift CLI', () => {
     );
   });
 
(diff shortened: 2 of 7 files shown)


facebook-github-bot-3:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

fkling:
Interesting... what would be a real use case for this feature?

facebook-github-bot-1:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

typeetfunc:
@fkling oh sorry...
Motivation: many kinds of refactoring include split big one file to many files. With ability to return many files will be able to make codemods such as:
 - `codemod-split-utils-file` - I will be split file with many exported functions to many files according the rule "one file - one function". Source file will used as entry point and contatins only `export from` statements.
 - `codemod-split-components-to-files` - It is the same as the previous but for React component
 - `codemod-export-complex-flow-type` - It move big and complex type annotations in separate file
 Also I want add to [codemod-proptypes-to-flow](https://github.com/billyvg/codemod-proptypes-to-flow) option for place generated annotation in separate file(something like `Component.js.flow` for `Component.jsx`).

fkling:
Overall, this makes sense to me, but I would be curious to know what others think about it.

Also, should it be possible to delete the original file?

typeetfunc:
>Also, should it be possible to delete the original file?

@fkling I think that correctly deleting a file is impossible in most cases.
For example, in `codemod-split-utils-file` you can not drop original file correctly because you will need to rewrite all imports of utils in a project. I.e. for simple and correctly delete original file `jscodeshift` must have feature for multi-round transformation then it will be possible to correctly rewrite all imports in a second round. This is a very complex feature and very different from my proposal.

AsaAyers:
I found this issue while looking to see if I could inline one file into another. I have a set of files with a bunch of `import * as whatever from './otherfile'` and I don't think `otherfile` needs to exist.

I was also looking to see if I could rename files because I have a project that follows:

`some-module/index.js` simply imports and re-exports `some-module/some-module.js`. So everything is simplified by renaming `some-module/some-module.js` to `some-module.js` and deleting the `index.js` file . I just have to transform any relative paths it might be importing from because it moved up 1 level.

gcazaciuc:
This is definetely something extremely usefull. Indeed there are many kinds of transforms that i would like to implement that require something like this. Some ideas where it would be usefull:

1. Separate React components into presentation & container components automatically based on dispatches made by dumb component.
2. Extract state from a React component into a container component

The only suggestion i have is maybe we could make the API a little simpler, eg return from the transform an object mapping file paths to AST like so:
```js
return {
     'file1': ast1,
     'file2': ast2
}
```
and have the worker create the specified files if they don't exist already.

PR #116, opened 2016-06-16
Upgrade recast to ^0.11.8

keyanzhang:
Fixed the issue of not printing return type in arrow function expressions (benjamn/recast#289). Thanks!


fkling:
Since this is a patch version update, all you need to do is reinstall jscodeshift I think. Updating the version here should not be necessary (of course we'll still do it).

PR #119, opened 2016-06-18
Add the possibility to select different parsers (closes #117)

### LINKED ISSUE #117: Allow custom parsers to be used by jscodeshift
The parser that is currently used, Babel v5, is outdated and doesn't understand the new flow type annotations.

However, simply upgrading to Babel / babylon v6 could break existing codemods, since v6 produces a slightly different AST.

That's why we make the parser _configurable_. By default jscodeshift will provide support for **Babel v5**, **babylon v6** and **flow**. The parser can be set either via the command line (e.g. `--parser=flow`) or can be specified by the transform (e.g. `export const parser = 'flow'`). Instead of selecting a predefined parser, the transform can also provide their own (e.g. `export const parser = someOtherParser;`), as long as it is compatible with recast (the selected parser is passed to recast).

I'm almost was done with this and just wanted to inform you that this is being worked on.

fkling:
jscodeshift uses Babel v5 to parse source files. Since Babel v5 doesn't get updated anymore, jscodeshift is unable to parse files that contain more modern flow type annotation.

To solve this, jscodeshift now supports three parsers: Babel v5, babylon and flow. The parser to use can be chosen by either

  - passing the command line option `parser`: `--parser=babelv5`,
    `--parser=babylon`, `--parser=flow`.
  - having the transform file export `parser`, with either the values
    `"babelv5"`, `"babylon"` or `"flow` to choose one of the built-int
    parsers, or export a parser object (e.g. `exports.parser = require('flow=parser'`)

    This allows transform to specify their own parser (as long as it is compatible with ESTree / recast), making them a bit more independent from jscodeshifts internals.

In addition to this change, `jscodeshift` now accepts a second argument, `options` which is directly parsed passed to `recast.parse`. So if you use `jscodeshift` programmatically , you can now pass your own parser with

```
jscodeshift(source, {parser: require('myParser')})
```

**Note:** This is *not* a breaking change. Babel v5 still stays the default parser, but we will likely use another parser (babylon or flow) as default in the future.

---

As another effect, I'm removing jscodeshift's own build step. Node v4 almost supports everything we want, I just removed the few occurrences of spread and destructuring. Using babel as part of the tool itself and as build tool is a but confusing.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,5 +9,8 @@
   },
   "env": {
     "node": true
+  },
+  "globals": {
+    "Promise": true
   }
 }
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,8 @@
 /src/
 /sample/
 .gitignore
+.eslintrc
+.eslintrc.yaml
 .jshintrc
 .module-cache
 __tests__
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,5 @@ sudo: false
 language: node_js
 node_js:
   - 4
+  - 5
   - stable
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ Options:
    --ignore-config FILE        Ignore files if they match patterns sourced from a configuration file (e.g., a .gitignore)
    --run-in-band               Run serially in the current process  [false]
    -s, --silent                No output  [false]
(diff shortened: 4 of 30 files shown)


fkling:
Thanks @benjamn! I'm curious to see how well the flow parser works...

PR #122, opened 2016-06-21
Upgrade lodash to v4

keyanzhang:
lodash v3 has an issue that causes `TypeError: Cannot read property 'prototype' of undefined` when we set Jest's `testEnvironment` to `node`. This has been fixed in v4 (lodash/lodash#223).

Thanks for reviewing! cc @cpojer


fkling:
Released as v0.3.23 .

keyanzhang:
Thanks @fkling!

PR #125, opened 2016-07-03
Fix renameTo to ignore non-variable identifiers

robz:
The old code to filter out member expression properties was broken (should be no ! before the computed check), and there are also other cases where an identifier is not a variable name.


fkling:
Great, thank you!

PR #128, opened 2016-07-06
Include stats and timing information in node API

iamdustan:
This allows for jscodeshift to be used first class for doing simple static analysis reporting.

For example I am using this change to detect and print out `process.env[VARIABLE]`s accessed and track our analytics calls to automate documenting what has been implemented and what still needs to be implemented.


fkling:
Makes sense to me, thank you!

Issue #126, opened 2016-07-04
Request: Support codemods published to npm


DrewML:
Right now, there are 2 ways that a developer can run a codemod:

1. Copy the codemod to their local machine, and pass the file path in with `-t`
2. Provide a URI to a remote `.js` file (with the caveat that the script must be entirely self contained).

I'd like to propose a 3rd option: being able to `npm install -g` a codemod.

I can think of 2 different options for how this would work.

1. A developer passes in the name of the globally-installed transform (either with `-t` or a new flag)
2. `jscodeshift` can look for all globally-installed modules matching a specific prefix, and present the developer with a list of available transforms when one is not explicitly provided on the command line (this is how the `yo` CLI works for `yeoman`).

I think the major benefit here would be that it then becomes easier to author + distribute a codemod that is not entirely self-contained. Right now, if I wanted to publish a codemod that uses `require` for modules other than node built-ins, a consumer of my codemod needs to:

1. Clone the repository
2. Run `npm install`
3. Navigate to their project
4. Specify the exact path to my codemod's entry point with the cloned repository

This is work I'd definitely be willing to contribute, but wanted to elicit feedback before taking that work on.

relekang:
I think it would be great if it was easier to consume codemods. I like both options.

Another option worth considering is custom cli. I did this with [my repo of codemods](https://github.com/relekang/codemods#usage). A benefit with this approach is that jscodeshift becomes a dependency of the codemod and thus the user knows that is it tested with the jscodeshift version they use it with. jscodeshift could have a helper which makes it a bit easier to generate the cli wrapper as well as keeping it in sync with jscodeshift cli options.

DrewML:
> Another option worth considering is custom cli.

That would be cool. [ava-codemods](https://github.com/avajs/ava-codemods) does a similar thing (wraps `jscodeshift` with a different CLI. While it's technically possible now, the API you can use isn't documented, which makes it less-obvious to newer consumers. Probably worth a separate issue on this repo.

fkling:
Sorry for the late response, I've been on vacation the last two weeks.

I like both ideas.

>  jscodeshift could have a helper which makes it a bit easier to generate the cli wrapper as well as keeping it in sync with jscodeshift cli options.
Definitely. A custom CLI has the advantage that additional preprocessing and configuration could happen before the codemod is run.

> A benefit with this approach is that jscodeshift becomes a dependency of the codemod and thus the user knows that is it tested with the jscodeshift version they use it with.

[...]

PR #133, opened 2016-07-22
Add eslint and babel-eslint to devDeps

DrewML:
Thanks for including a lint config!

I have my editor setup to always use a packages local copy of `ESLint` to prevent possible version /config problems. Noticed it failing out when working on #126.

Any problems with adding them? I can work-around it if not.


fkling:
Thank you! No problem at all, no idea why it isn't in there yet :)

PR #138, opened 2016-07-30
Custom cli helper

### LINKED ISSUE #132: Custom cli helper
In #126 it was proposed to add a helper for generating custom cli for codemod collections. 

The reasoning mentioned there:
> Another option worth considering is custom cli. I did this with [my repo of codemods][rcm]. A benefit with this approach is that jscodeshift becomes a dependency of the codemod and thus the user knows that is it tested with the jscodeshift version they use it with. jscodeshift could have a helper which makes it a bit easier to generate the cli wrapper as well as keeping it in sync with jscodeshift cli options.

As I mentioned in #126 I will have time to start implement this next week.

I was thinking that the usage would be that the bin script of a codemod collection would be something like this:

```javascript
var path = require('path');
var cli = require('jscodeshift').cli;

cli({
  list: function list() {
    return [{name: 'template-strings', description: '...'}];
  },
  resolve: function resolve(name) {
    return path.join(__dirname, 'transforms', name + '.js');
  }
});
```

[rcm]: https://github.com/relekang/codemods#usage

relekang:
Makes it possible to create cli interfaces for codemod collections. See #132 
### Example usage

``` javascript
#!/usr/bin/env node

const resolve = require('path').resolve;
const cli = require('jscodeshift').cli;

cli({
  name: 'Custom CLI',
  packageVersion: '2.3.0',
  resolve(name) {
    return resolve(__dirname, `./transforms/${name}.js`);
  },
});
```
#### `--help` output

```
Usage: Custom CLI <transform> <path>... [options]

transform     The transformer to use
path          Files or directory to transform

Options:
   -c, --cpus             (all by default) Determines the number of processes started.
   -v, --verbose          Show more information about the transform process  [0]
   -d, --dry              Dry run (no changes are made to files)
   -p, --print            Print output, useful for development
   --babel                Apply Babel to transform files  [true]
   --extensions           File extensions the transform file should be applied to  [js]
   --ignore-pattern       Ignore files that match a provided glob expression
   --ignore-config FILE   Ignore files if they match patterns sourced from a configuration file (e.g., a .gitignore)
   --run-in-band          Run serially in the current process  [false]
   -s, --silent           No output  [false]
   --parser               The parser to use for parsing your source files (babel | babylon | flow)  [babel]
   --version              print version and exit
```
#### `--version` ouput

```
Custom CLI: 2.3.0
 - jscodeshift: 0.3.26
 - babel: 5.8.38
 - babylon: 6.8.4
 - flow: 0.28.0
 - recast: 0.11.10
```

Fixes #132

### DIFF
--- a/src/__tests__/custom-cli-test.js
+++ b/src/__tests__/custom-cli-test.js
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2015-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+'use strict';
+
+/*global jest, describe, it, expect*/
+
+jest.autoMockOff();
+
+const cp = require('child_process');
+const resolve = require('path').resolve;
+const cli = resolve(__dirname, 'custom-cli');
+
+function execFile(file, args) {
+  return new Promise((resolve, reject) => {
+    cp.execFile(file, args, (error, output) => {
+      if (error) return reject(error);
+
+      resolve(output);
+    });
+  });
+}
+
+describe('Custom CLI', function() {
+  it('should show help when --help is passed', function() {
+    return execFile(cli, ['--help'])
+      .then(function (result) {
+        expect(result).toContain('Usage: Custom CLI <transform> <path>');
+      })
+  });
+
+  it('should resolve transform', function() {
+    return execFile(cli, ['noop', resolve(__dirname, 'custom-cli-test.js')])
+      .then(function (result) {
+        expect(result).toContain('Using transform noop');
+        expect(result).toContain('Processing 1 files...');
+      })
+  });
+
+  it('should fail when transform is not resolved', function() {
+    let catched = false;
+    return execFile(cli, ['non-existing-codemod', resolve(__dirname, 'custom-cli-test.js')])
+      .catch(function (error) {
+        catched = true;
+        expect(error.toString()).toContain('The transform non-existing-codemod does not exist.')
+      })
+      .then(() => {
+        expect(catched).toBe(true);
+      });
+  });
+});
--- a/src/__tests__/custom-cli-test.js
+++ b/src/__tests__/custom-cli-test.js
@@ -47,14 +47,14 @@ describe('Custom CLI', function() {
   });
 
   it('should fail when transform is not resolved', function() {
-    let catched = false;
+    let caught = false;
     return execFile(cli, ['non-existing-codemod', resolve(__dirname, 'custom-cli-test.js')])
       .catch(function (error) {
-        catched = true;
+        caught = true;
         expect(error.toString()).toContain('The transform non-existing-codemod does not exist.')
       })
       .then(() => {
-        expect(catched).toBe(true);
+        expect(caught).toBe(true);
       });
   });
 });
--- a/src/__tests__/custom-cli-test.js
+++ b/src/__tests__/custom-cli-test.js
@@ -29,6 +29,8 @@ function execFile(file, args) {
 }
 
 describe('Custom CLI', function() {
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
+
   it('should show help when --help is passed', function() {
     return execFile(cli, ['--help'])
       .then(function (result) {
--- a/src/cli.js
+++ b/src/cli.js
@@ -10,7 +10,6 @@
 
  /* eslint-disable no-console */
 
-const _ = require('lodash');
 const fs = require('fs');
 
 const pkg = require('../package.json');
@@ -145,7 +144,7 @@ function cliGenerator(cliOptions) {
 
   const args = require('nomnom')
     .script(name)
-    .options(_.assign({}, defaults, {version, transform, path}))
+    .options(Object.assign({}, defaults, {version, transform, path}))
     .parse();
 
   const transformerPath = resolve(args.transform);
@@ -160,7 +159,7 @@ function cliGenerator(cliOptions) {
     Runner.run(
       transformerPath,
       args.path,
-      _.assign({}, options, args)
+      Object.assign({}, options, args)
     );
   });
 }
--- a/src/cli.js
+++ b/src/cli.js
@@ -8,7 +8,10 @@
  *
  */
 
+ /* eslint-disable no-console */
+
 const _ = require('lodash');
+const fs = require('fs');
 
 const pkg = require('../package.json');
 const Runner = require('../dist/Runner.js');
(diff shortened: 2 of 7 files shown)


relekang:
I added [some documentation](https://github.com/relekang/jscodeshift/blob/custom-cli-helper/README.md#custom-cli-helper-for-codemod-collections).

fkling:
Sorry that I haven't taken a look at this yet. I will do that this week!

src/cli.js
```diff
@@ -0,0 +1,168 @@
...
+      console.error(`The transform ${args.transform} does not exist.`);
+      process.exit(1);
+    }
+
+    console.log(`Using transform ${args.transform}.`);
+
+    Runner.run(
+      transformerPath,
+      args.path,
+      _.assign({}, options, args)
```
DrewML: `jscodeshift` [only supports node 4 and up](https://github.com/facebook/jscodeshift/blob/master/package.json#L21), so you could just use the builtin `Object.assign` here

DrewML:
Thanks for doing this!

I don't want to discourage your efforts at all (this isn't even my project!), but for the sake of discussion, I think there may be room here to provide a solution that will be more easily consumed by users of published codemod collections.

I've authored quite a few codemods at my job, and I've always found that people tend to be less enthused about using them when they realize they have to figure out the right path (or in this PR's case, name) of the transform, along with whatever other options they may need to pass.

One option I've seen that I _really_ liked was what the `AVA` team did with the [ava-codemods](https://github.com/avajs/ava-codemods) project. You can run the executable without reading any docs or `--help` text, and it basically guides you through a set of steps, that includes choosing from a list of predefined codemods.

Another benefit to the "guided" setup is that authors of codemods will typically already know the answer to some questions, and can answer them on behalf of the user. As an example, if I write a codemod that relies on the AST structure provided by the `Flow` parser, I'd like to be able to just set the parser automagically, and not make a user of my CLI have to guess at it.

I think it would be really cool if, rather than providing a way to customize a few pieces of the `jscodeshift` CLI, we could work to expose enough APIs that would make creating a solution like that much simpler.

Thoughts?

relekang:
First of all, it is not discouraging. Feedback is good so thank you @DrewML for taking the time to write it :)

I also really like the approach AVA has taken to this. However, I think there are different usage pattern in fixing breaking changes in upgrading a library and all the other things we can do with codemods. I also think that a guided version would be good and I think this PR is a step in the right direction for making that happening.

My initial thought with the current approach in the PR was to create a second PR that added `resolveList` in the same manner as `resolve` and if the user did not specify a codemod name we could list it out. After what you said, I'm thinking it could use that information to give the user a list that they can choose codemods from. I'm happy to work on extending this in that direction, but it would be awesome to land this one first.

Providing a way to set options for a given codemod collection is just the matter of adding an options field to the cli object. Which I think sounds like a good idea.

fkling:
Thanks for starting a discussion @DrewML! The reason why I was hesitant to make any move on this was that *I'm* not the one using an API like this (i.e. I'm not writing codemod collections), so I don't feel like the right person to make a decision about this.

My main worry is around having to make breaking changes if the API doesn't allow people to do what they want to do.

However, it seems the changes here provide a basic foundation, that gives us enough possibilities to build upon and expand it without having to make breaking changes.

Would still like to get more feedback from people who would actually use this API.

DrewML:
Thanks for the replies @fkling and @relekang. I definitely agree this is a great start!

> Would still like to get more feedback from people who would actually use this API.

That's a great idea. @jfmengels + @jamestalmage, since you two did a good chunk of the work on `ava-codemods`, do y'all have any input on this API based on your experience so far?

src/__tests__/custom-cli-test.js
```diff
@@ -0,0 +1,60 @@
...
+        expect(result).toContain('Using transform noop');
+        expect(result).toContain('Processing 1 files...');
+      })
+  });
+
+  it('should fail when transform is not resolved', function() {
+    let catched = false;
+    return execFile(cli, ['non-existing-codemod', resolve(__dirname, 'custom-cli-test.js')])
+      .catch(function (error) {
+        catched = true;
```
jfmengels: catched --> caught?

jfmengels:
After writing the CLI for ava-codemods, I actually extracted it and created https://github.com/jfmengels/lib-upgrader/, which I use in https://github.com/jfmengels/lodash-codemods/. I never re-integrated it into ava-codemods, but that's scheduled to be done... at some point.

That said, I'd like to point out a few things:
In `lib-upgrader` / `ava-codemods`, we ask questions:
  - whether you want to upgrade your codebase from one version of a library (AVA, Lodash, ...) to another, and if so we ask from which to which (currently done).
  - Or if you rather want to migrate from a different library to this library (Tape --> AVA, underscore --> Lodash). At the moment, this is done via a different executable for AVA, but that could change. It's also not done for Lodash due to a lack of codemods for it, but both projects could use it at some point.

According to the results, we might want to run several codemods. For instance, to migrate from v1.0.0 to v3.0.0, we'll apply the codemods for v1.0.0 to v2.0.0 and v2.0.0 to v3.0.0. At the moment, we spawn the runner manually several times.

(
Speaking about that, but this is probably a whole different issue: Spawning the runner is actually really slow.
I notice that mostly in lodash-codemods where I have 4-5 codemods to switch from v3 to v4 (each codemod takes care of different concerns of the migration, in order to have manageable codemod files and nice tests). And the way it is run is
- Run codemod 1 on all the files
- Then run codemod 2 on all the files
- Then run codemod 3 on all the files
- ...
It would be a lot faster I think if you could run multiple transforms in the same command and do it like
- For every file
  - Run codemod 1
  - Run codemod 2
  - Run codemod 3
  - ...

)

The way this CLI is made, you can only run one codemod and not ask questions. That is fine if you only have one codemod to run (like: change all my tabs to spaces) and the user is properly informed of the command he must run. For this, the API looks really small and therefore I think it's pretty nice. :+1: for that use case.

But this would not be enough IMO in lodash-codemods and ava-codemods, as the experience for the user is much better IMO when using `lib-upgrader`.

@relekang Also, for the provided example, you should require `package.json` and read the version from there. Easier to keep up to date.

I hope this helps. Let me know if you want to discuss some of the off-topic items in a new issue :)

DrewML:
@jfmengels Thanks for the input!

> It would be a lot faster I think if you could run multiple transforms in the same command

This is being discussed in https://github.com/facebook/jscodeshift/issues/148

relekang:
> My main worry is around having to make breaking changes if the API doesn't allow people to do what they want to do.

Isn't jscodeshift a tool that could help us handle that :v: but on the serious side, I see you concern and I agree that we should get as much feedback as possible on this from.

@jfmengels Thanks for sharing. 

About the lib-upgrader use case. I see it as another specific use case that deserves its own "helper" and your library is cool. That said, it would be nice to be able to ask questions. 

> Also, for the provided example, you should require package.json and read the version from there. Easier to keep up to date.

👍 Updated it

src/__tests__/custom-cli-test.js
relekang: 👍 Changed it

src/cli.js
relekang: Nice, changed it 👍

PR #142, opened 2016-08-13
Add .editorconfig to root

DrewML:
Used [editorconfig-tools](https://www.npmjs.com/package/editorconfig-tools) to generate a config inferred from existing code in the repo.

Makes keeping the code style consistent a bit easier for contributors that have default editor settings for different preferences.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


fkling:
Sure, why not 😄

Issue #150, opened 2016-09-18
Adding a .length getter instead of .size()


vjeux:
It's really confusing that you cannot use the regular `.length` and have to use `.size()`. All the environments where we would want to run this support getters so it would be great so support!

PR #151, opened 2016-09-18
Deprecate Collection#size(), add Collection#length

DrewML:
Implements #150 (assuming it is wanted).

I was going to use [`util.deprecate`](https://nodejs.org/api/util.html#util_util_deprecate_function_string), but cannot do that without using the class fields proposal, so I'm just writing a warning to `stderr` atm.

`EditorConfig` setup in Atom killed some extra whitespace. Happy to revert if requested.

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -114,6 +114,15 @@ class Collection {
   }
 
   /**
+   * Returns the number of elements in this collection.
+   *
+   * @return {number}
+   */
+  get length() {
+    return this.__paths.length;
+  }
+
+  /**
    * Returns an array of AST nodes in this collection.
    *
    * @return {Array}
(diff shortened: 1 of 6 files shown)


vjeux:
<3

Any thoughts @fkling?

src/Collection.js
```diff
@@ -110,6 +110,16 @@ class Collection {
    * @return {number}
    */
   size() {
+    process.stderr.write('Collection#size() is deprecated. Use Collection#length instead');
```
vjeux: I don't see much harm supporting both. I would just add the other one and not deprecate this one.

fkling:
👍 I also think there is no need to add a deprecation warning. Rest looks good to me 😄

DrewML:
Removed deprecation. @fkling how do you want me to handle tests for `size()`. I can add an assertion on `size()` in every test that we check `length` in, or I can just add 1 explicit test to ensure that `size()` and `length` return the same value.

fkling:
I'm actually OK with as it is. No need to add tests back.

PR #152, opened 2016-09-21
Fix replaceWith to retain correlated comments

zackargyle:
This fixes the bug that requires this recipe: [retain-first-comment](https://github.com/facebook/jscodeshift/blob/master/recipes/retain-first-comment.md). Essentially if the initial expression is replaced, the top-of-file comments are removed by default. An eslint autofixer had this issue, and a [codemod](https://github.com/zackargyle/codemods/blob/master/underscore-to-native.js) I recently wrote had the same issue until I fixed it, but it seems like this should be default behavior. I've verified that it works on my codemod.

### DIFF
--- a/src/collections/Node.js
+++ b/src/collections/Node.js
@@ -136,7 +136,11 @@ var mutationMethods = {
     return this.forEach(function(path, i) {
       var newNodes =
         (typeof nodes === 'function') ? nodes.call(path, path, i) : nodes;
+      var comments = path.value.comments;
+      var loc = path.value.loc;
       path.replace.apply(path, toArray(newNodes));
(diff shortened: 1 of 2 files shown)


DrewML:
I could be wrong, but I'm not sure that `jscodeshift` is the right place to fix this (vs `recast` itself). There are several different ways (besides `replaceWith`) that the AST could be manipulated, that will result in the first comment being wiped out when that isn't the expected behavior.

[Contrived Example](http://astexplorer.net/#/AfUZM12THf)

zackargyle:
@DrewML you're absolutely right. I just through this together quickly and didn't think about it. Should definitely be a recast PR. My bad!

PR #153, opened 2016-09-24
Add j shorthand

vjeux:
This way we can write

```js
export default function transformer(file, { j }) {
  return j(file.source).toSource();
}
```

and not need to have the annoying local variable

```js
export default function transformer(file, api) {
  const j = api.jscodeshift;

  return j(file.source).toSource();
}
```


vjeux:
Thanks!

PR #154, opened 2016-09-24
Update to jest 15

vjeux:
As I wrote a test, I didn't get any of the fancy output of jest 15 so I upgraded. 

It makes the tests go from 56s to 17s :) https://github.com/facebook/jscodeshift/pull/155#issuecomment-249397496

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "optional": ["runtime"]
-}
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
   "devDependencies": {
     "babel-eslint": "^6.1.2",
     "eslint": "^3.1.1",
-    "jest-cli": "^12.0.0",
     "jsdoc": "^3.4.0",
+    "jest": "^15.1.1",
     "mkdirp": "^0.5.1"
   },
   "jest": {
@@ -56,9 +56,6 @@
       "src",
       "bin",
       "sample"
-    ],
-    "unmockedModulePathPatterns": [
-      "node_modules/"
     ]
   }
 }
--- a/src/__tests__/Collection-test.js
+++ b/src/__tests__/Collection-test.js
@@ -10,19 +10,25 @@
 
 'use strict';
 
-jest.autoMockOff();
-
-var recast = require('recast');
-var Collection = require('../Collection');
-
-var NodePath = recast.types.NodePath;
-var types = recast.types.namedTypes;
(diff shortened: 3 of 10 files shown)


.babelrc
```diff
@@ -1,3 +0,0 @@
-{
-  "optional": ["runtime"]
```
vjeux: I'm not really sure why this was there but jest complained

>    ReferenceError: [BABEL] /Users/vjeux/random/jscodeshift/bin/__tests__/jscodeshift-test.js: Using removed Babel 5 option: /Users/vjeux/random/jscodeshift/.babelrc.optional - Put the specific transforms you want in the `plugins` option
>      at Logger.error (node_modules/babel-jest/node_modules/babel-core/lib/transformation/file/logger.js:41:11)

package.json
```diff
@@ -47,18 +47,15 @@
...
+    "jest": "^15.1.1",
     "mkdirp": "^0.5.1"
   },
   "jest": {
     "testPathDirs": [
       "src",
       "bin",
       "sample"
-    ],
-    "unmockedModulePathPatterns": [
```
vjeux: automock was already manually turned off on every single file so this didn't do anything :)

src/__tests__/Collection-test.js
```diff
@@ -10,19 +10,25 @@
...
 describe('Collection API', function() {
   var nodes;
+  var Collection;
+  var recast;
+  var NodePath;
+  var types;
+  var b;
 
   beforeEach(function() {
+    jest.resetModuleRegistry();
```
vjeux: jest 15 no longer calls this by default. This is the only file where it is actually needed but I figured I would add it on the other files as well since this pattern is used so that modules are expected to be isolated.

src/__tests__/Worker-test.js
```diff
@@ -10,8 +10,6 @@
 
 'use strict';
 
-jest.autoMockOff();
```
vjeux: <3

.babelrc
fkling: I actually think that we don't need that file at all anymore, since jscodeshift doesn't have a build process anymore since 327f985598784e1e91615d78237ece490995d165.

vjeux: Cool, let's ship it then!

hzoo: "optional": ["runtime"] would use babel-runtime in babel 5

PR #155, opened 2016-09-25
Add concurrency to jscodeshift-test to make runtime go from 56s to 34s

vjeux:
Jest 16 will have it.concurrent which lets you run all the promises at once and not serially. It turns out that the implementation is just a few lines of js that we can embed while it gets released.

There were two issues to make them actually concurrent:
1) sources was a variable inside of describe that shouldn't be shared anymore
2) createTempFileWith when passed the same filename would result in the same path which conflicted with other tests. Now it's unique (as it should have been to begin with)


vjeux:
node 4, jest 12, baseline: 56s

screen shot 2016-09-24 at 6 35 51 pm

node 4, jest 12, concurrent: 34s

screen shot 2016-09-24 at 6 35 40 pm

node 6, jest 15, baseline: 17s

screen shot 2016-09-24 at 6 40 25 pm

node 6, jest 15, concurrent: 9s

screen shot 2016-09-24 at 6 39 31 pm

vjeux:
Upgrading to jest 15 (and node 6 which has a small impact) make the tests go from 56s to 17s! In addition, this change makes the entire test suite run in 10s from around a minute :)

One unfortunate side effect is that because this is the test that takes the biggest amount of time and now we make it CPU bound, it messes up the timing numbers of all the other tests which are starved for CPU time.

vjeux:
Travis on node 4 is failing and I can't figure out why. I refactored the code to create a temp folder and have all the files in there but it still doesn't work.

With jest 15, the total runtime is only 15s so putting it to 10s isn't worth me spending more time on it :)

PR #156, opened 2016-09-29
Atomically write transformed JS files back to the filesystem

alangpierce:
Instead of directly overwriting each JS file with its new contents, the worker
code now writes the new contents to a temporary file and then does a filesystem
move operation to overwrite the old file. This avoids the previous behavior
where the file had invalid contents while it was being written (because
`writeFile` is not atomic).

This fixes a race condition I was seeing in a script I was running on a large codebase:
https://github.com/alangpierce/bulk-decaffeinate/blob/master/jscodeshift-scripts/fix-imports.js

The script fixes import statements across the codebase by reading the exports
of the referenced files (using `readFileSync` and using jscodeshift to parse
the other file). Because the other files were also being transformed by
jscodeshift, sometimes (nondeterminsitically) some imports would fail to
convert because the referenced file wouldn't be valid. After applying this fix
and re-running the script on my codebase, it seems to completely fix the
problem.

### DIFF
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -146,13 +146,21 @@ function run(data) {
             console.log(out); // eslint-disable-line no-console
           }
           if (!options.dry) {
-            fs.writeFile(file, out, function(err) {
+            const tmpFile = `${file}.tmp.${Math.floor(Math.random() * 1000000)}`;
+            fs.writeFile(tmpFile, out, function(err) {
               if (err) {
                 updateStatus('error', file, 'File writer error: ' + err);
+                callback();
               } else {
-                updateStatus('ok', file);
+                fs.rename(tmpFile, file, function(err) {
+                  if (err) {
+                    updateStatus('error', file, 'File move error: ' + err);
+                  } else {
+                    updateStatus('ok', file);
+                  }
+                  callback();
+                });
               }
-              callback();
             });
           } else {
             updateStatus('ok', file);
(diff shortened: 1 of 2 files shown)


src/Worker.js
```diff
@@ -146,13 +146,21 @@ function run(data) {
             console.log(out); // eslint-disable-line no-console
           }
           if (!options.dry) {
-            fs.writeFile(file, out, function(err) {
+            const tmpFile = `${file}.tmp.${Math.floor(Math.random() * 1000000)}`;
```
wbinnssmith: what do you think about using `os.tmpdir` with a stricter permission (0600) or even just the [`tmp`](https://www.npmjs.com/package/tmp) module's `file` method so we don't clutter the pwd in case of failure?

Currently if the write succeeds but the rename does not the randomly named transformed file will be left behind.

wbinnssmith:
love this!

alangpierce:
Sorry for the delay in getting to this. One issue I ran into with this change and other temp file approaches is that I want to preserve permissions on the original file. In particular, if the original file is an executable script, I want it to stay that way. It doesn't look like there's any way to preserve the destination file permissions on `fs.rename`.

The [write-file-atomic](https://www.npmjs.com/package/write-file-atomic) package seems to handle temp file creation (although not using `os.tmpdir`), keeping the permissions correct on the target file, and cleaning up the tmp file if there was any error, so I just updated the PR to use that (which also makes the code change much simpler). Let me know what you think!

PR #162, opened 2016-10-14
Add ability to use jsx to represent ASTs

zackargyle:
Before I continue to far down this path, I wanted to present the idea. Essentially this PR would allow you to represent ASTs in codemods using JSX syntax. It's like the "enzyme" of codemods. It provides a much simple search mechanism and simple APIs for replacing and updating ASTs. Part of the motivation is that I've found it difficult to map the AST itself to matching algorithms, and impossible to remember argument order when creating new trees. Here are a few examples to show you how it works currently.

This replaces all '_' identifiers with '$'.

``` js
// file.source = "const result = _.map([1,2,3], () => {});"
return jsx(file.source)
    .find(<Identifier name="_" />)
    .setProp('name', '$')
    .toSource();
```

This swaps out all 'i18n' identifiers with a 'i18n._' member expression

``` js
// file.source = "const text = i18n('Some text');"
return jsx(file.source)
    .find(<Identifier name="i18n" />)
    .replaceWith(node => (
        <MemberExpression>
            <Identifier name="i18n" />
            <Identifier name="_" />
        </MemberExpression>
    )).toSource();
```

You can see more examples of usage in `AstWrapper-test.js`.

In general it feels like a good direction. Some type of higher-level abstraction to manage transformations. Hardest part is that with my current design we need a `class` for each AST type. I have Identifier, Literal, MemberExpression, CallExpression, and VariableDeclarator in the codebase right now.

Let me know your thoughts before I keep hacking on this!


kevinbarabash:
I feel your pain with respect to argument order.  The thing that concerns me with this approach is that this doesn't actually solve the argument order issue.  In the example I don't actually know if this renders to `i18n._` or `_.i18n`.  

An alternative that might help with argument order is using object literals as arguments, e.g.
```
b.memberExpression({
    object: b.identifier('i18n'),
    property: b.identifier('i18n'),
});
```

Another possible solution would be to add jsdocs or have a .d.ts file that has the parameter order to the current builder functions.

zackargyle:
See I disagree. The object/property is unintuitive to me, whereas child1/child2 seems much more intuitive. The main benefit I think is still in the "find" function using JSX. So much easier than grabbing the top level node and digging down manually into its children to verify.

cpojer:
Like we discussed in private, I love this @zackargyle. I've wanted something like this for ages.

kevinbarabash:
@zackargyle I definitely like that `find` here can filter by `name`.  Can you pass other nodes to `find`?

zackargyle:
@kevinbarabash Absolutely. Currently it takes any supported type, with empty props/children used as wildcards. So something like this would find all underscore method calls (`_.map(...)`, `_.throttle(...)`, ...).

```
const Underscore = () => (
    <CallExpression>
        <MemberExpression>
            <Identifier name="_" />
            <Identifier />
        </MemberExpression>
    </CallExpression>
);
wrapper.find(<Underscore />)
```

fkling:
Thank you for starting this! I very much agree that we need a simpler way of specifying search patterns. I will first talk about the general idea of improving search patterns and then comment on your specific approach.

First a quick note to your second example: We already have a mechanism to create replacement nodes that is more convenient IMO: [templates](https://github.com/facebook/jscodeshift/blob/master/src/template.js):

```js
// file.source = "const text = i18n('Some text');"
return jsx(file.source)
    .find(<Identifier name="i18n" />)
    .replaceWith(j.template.expression`i18n._`)
    )).toSource();
```

This also brings me to my next point: While JSX makes it a bit nicer to write these patterns (and can give you named parameters, which is great), I wonder if it is the right level of abstraction here. If we really want to make it easier to search for a piece of code, we should enable people to actually write the code snippet to search for.

For this to work we would need some kind of language that enables us to express intent within code snippets. E.g. `$$` could stand for an arbitrary expression / node, `$$Identifier` for a node of type *Identifier*.

Applying that idea to your examples, the first one could look something like

```js
// file.source = "const result = _.map([1,2,3], () => {});"
return j(file.source)
    .find(match.expression`_`)
    .replaceWith(template.expression`$`)
    .toSource();
```

```js
// file.source = "const text = i18n('Some text');"
return jsx(file.source)
    .find`match.expression`i18n`)
    .replaceWith(template.expression`i18n._`)
    .toSource();
```

```js
const underscore = match.expression`_.$$()`;
wrapper.find(underscore)
```

---

However, I don't want to discourage you from going down the JSX route. The advantage of JSX (and the raw object literals) over this hypothetical matching language is that you can express more complex matching conditions more easily since you have access to raw JavaScript.

What I like about your approach is that you are building on top of jscodeshift's existing `.find` method, the `debug` method and the way you implement the matchers for the components.  

What I don't know like as much is that you are introducing a new function (`jsx`) and wrapper object (`ASTWrapper`). jscodeshift aims to provide a unified interface. While it is extensible, one should have to think which entry function to use and what different entry functions return.

Can you elaborate on why you added that wrapper?  
I believe we can make some changes to jscodeshift itself so that wrappers like this one wouldn't be necessary. For example, you can easily add the `debug` and `setProp` methods to all *Node*s via `registerMethods`:

```js
j.registerMethods(
  {
    debug() { ... },
    setProp() { ... },
  },
   j.Node
);
```

Overwriting `.find` to only accept a single parameter is not possible atm, but we can fix that 😃 . I propose to change `.find` to either accept _one_ or _two_ arguments. If only a single argument is passed, it will be equivalent as searching for all *Node*s:

```js
j(src).find(arg)
// equivalent to
j(src).find(arg.type || j.Node, arg)
```

The second argument can already either be a function or an pattern object, which is flexible enough.

I believe this should be enough to support your use case, but let me know if it doesn't.

With that change, your proposal *could* be implemented as a separate plugin instead of having to be part of core (happy to discuss this more). I adapted your code to work with changes I had in mind for `.find`, and this would be the result (I didn't include all the file): https://gist.github.com/fkling/a8a80faeb8f179c873cad167f7eb2e5f . Note that this doesn't require a wrapper anymore.

I could see jscodeshift and matchers have a similar relationship as testing frameworks and assertion libraries have: You can choose which one to use.

zackargyle:
Hey Felix,

Really appreciate the feedback! And I agree on many accounts, particularly the replaceWith. The part I was most excited about was the find functionality. Part of the reason I like the JSX approach is because it has such a familiar feel to it. Particularly when using something like astexplorer ;). The wildcard matching via JSX makes it really awesome. For example, how would I find all underscore CallExpressions using a string? Maybe something like `_.*(*)`? Parsing/matching that to an AST structure is going to be really rough.

As far as the AstWrapper, my thinking was really on having a clear separation of the two syntaxes. Combining them will make it much harder to maintain, and you can always eject from the ast wrapper with `wrapper.ast`. 

The plugin would be tough because we need a way to tell babel to use our particular pragma to parse jsx `[require('babel-plugin-transform-react-jsx'), { pragma: 'jsx.createElement' }],`. We could, of course, add another cli option, but that would get more complicated.

So I definitely like the idea of this being a part of core, we just gotta figure out exactly how we want the implementation done. If other people also don't like the separate `jsx(file.source)` syntax, we can change that up for sure.

@fkling

zackargyle:
As a side note, I just pushed some changes to the way I defined ast components. It should be a lot easier to write them now. We use a similar PropTypes structure that is used for component matching, and validation before convert to an AST.

```
const j = require('../../core');
const PropTypes = require('../PropTypes');

module.exports = class CallExpression {

  static propTypes = {
    arguments: PropTypes.array.isRequired,
    children: PropTypes.children(['callee']).isRequired,
  };

  static toAST = (props) => {
    const children = props.children;
    return j.callExpression(
      children[0].constructor.toAST(children[0].props),
      props.arguments
    );
  }

}
```

fkling:
> For example, how would I find all underscore CallExpressions using a string? Maybe something like `_.*(*)?`

That's what I meant with coming up with some kind of language. E.g. for this example it might look like 

```js
match.expression`_.$$($$*)`
```

which could internally be converted to `'_.$$($$_varargs)'` (which is valid JS) and then could be parsed into the pattern object 

```js
{
  type: 'CallExpression',
  callee: {type: 'MemberExpression', object: {name: '_'}}
}
```

Similarly, 

```js
match.expression\`_.$$($$+)`
```

(at least one argument) would be converted to


```js
{
  type: 'CallExpression',
  callee: {type: 'MemberExpression', object: {name: '_'}},
  arguments: a => a.length > 0,
}
```

Of course this has limitations, because whatever we convert the code pattern into must be valid JavaScript, but it will probably cover 90% of all use cases.

>  Parsing/matching that to an AST structure is going to be really rough.

We are already doing part of this with the template functionality. The basic idea is: Parse the code into an AST, find all identifiers that have special names and replace them with a different value.

In the case of the templates it's the AST nodes you pass in. In case of matchers it would be functions.

> As far as the AstWrapper, my thinking was really on having a clear separation of the two syntaxes.

I don't see why the syntax for *defining the search pattern* should have an impact on what type of collection you are working with. The idea behind jscodeshift is that there could potentially be many ways to specify which nodes to find, but you will get back the same type of result (a collection of paths).

jscodeshift provides a low level find / filter API that either accepts a function or a pattern object. *How* that function/pattern object is created doesn't matter.

```
+---------------+          +---------------+         +---------------+
|               |          |   function/   |         |               |
| PatternSyntax +--------->|pattern object |-------->|     find      |
|               |          |               |         |               |
+---------------+          +---------------+         +---------------+
```

Why should a collection that was "found" via `.find(<Identifier />)` be any different than using `.find('Identifier')`? That seems like it could create a lot of confusion.

> The plugin would be tough because we need a way to tell babel to use our particular pragma to parse jsx `[require('babel-plugin-transform-react-jsx'), { pragma: 'jsx.createElement' }]`,. We could, of course, add another cli option, but that would get more complicated.

See the [gist](https://gist.github.com/fkling/a8a80faeb8f179c873cad167f7eb2e5f) I created. Users would have to add the `@jsx` pragma to the transform, which I agree is not ideal, but isn't that bad either.

> So I definitely like the idea of this being a part of core, we just gotta figure out exactly how we want the implementation done. 

Whether this should be core or not is a discussion we should defer IMO. I think the more interesting aspect of this PR is to figure out a useful/flexible protocol for matchers and replacers. Then we can achieve loose coupling and maintaining every part individually wouldn't be an issue.

zackargyle:
I'm totally happy merging the two together so that we don't have a separate parse function. That should be a fairly easy change and I think you're right. I do, however, think it's good to have a separate AstWrapper. It's not a Collection, and the Collection would have to be completely overhauled to support something other than a list of paths. The JSX is really a hybrid of a path + filter. Having worked with something like Enzyme, a wrapper with prop based syntax feels really great to use.

As for the Regexy `.find()` style you're suggesting, it doesn't feel like the right direction. No one really likes regex, whereas people love the feel of JSX, and it feels fitting for a tree-based transformation tool. Everyone I've talked to loves the idea of using JSX for codemods; I'm curious what their thoughts would be about your regex solution.

I realize I'm saying things like "feels good" a lot, and that's on purpose. I really want writing codemods to feel good, feel familiar, and feel approachable.

zackargyle:
@fkling Just committed an update that removes the AstWrapper and merges everything into core. I think it was a good call, and cleaned things up quite a bit. Good call.

src/collections/Node.js
```diff
@@ -32,6 +35,9 @@ var traversalMethods = {
    * @return {Collection}
    */
   find: function(type, filter) {
+    if (type.__IS_JSX__) {
```
fkling: If this was changed to

```diff
+    if (!filter && !types[type]) {
+      filter = type;
+      type = filter.type || 'Node';
+    }
```

instead and move the `matchJSX`, then we don't need special treatment for JSX.

src/core.js
```diff
@@ -123,6 +123,25 @@ function use(plugin) {
...
+ * @augments core
+ * @static
+ * @param {}
+ */
+function createElement(Component, props, ...children) {
+  const finalProps = Object.assign(props || {}, { children });
+  const astNode = new Component(finalProps);
+  astNode.__IS_JSX__ = true;
+  astNode.props = finalProps;
+  return astNode;
```
fkling: If you call `matchJSX` here, then it doesn't have to get special treatment in `find`:

```js
const matcher = matchJSX(astNode);
matcher.type = astNode.constructor.name;
return matcher;
```

src/collections/Node.js
```diff
@@ -142,6 +148,9 @@ var mutationMethods = {
     return this.forEach(function(path, i) {
       var newNodes =
         (typeof nodes === 'function') ? nodes.call(path, path, i) : nodes;
+      if (newNodes.__IS_JSX__) {
```
fkling: If this was changed to:

```js
let newNodes = nodes;
if (nodes && typeof nodes.toAST === 'function') {
  newNodes = nodes.toAST()
} else if (typeof nodes === 'function') {
  newNodes = nodes.call(path, path, i);
}
```

then *any* value that implements a `toAST` method could be passed as replacement, making it possible to define custom replacers.

Same below.

src/core.js
fkling: Similarly, if you assign `toAST` here, we don't need special treatment in replacement methods (assuming the corresponding changes mentioned there):

```js
matcher.toAST = () => jsxToAST(astNode)
```

jedwards1211:
@fkling 
> If we really want to make it easier to search for a piece of code, we should enable people to actually write the code snippet to search for.

This must be why the templates handle expressions the way they do, right?
For example I noticed that
```js
require('jscodeshift').template.statement`import * as ${() => true} from 'react'`
```
manages to get my function in place of `node.specifiers[0].local` with some clever substitution:
```
Node {
  type: 'ImportDeclaration',
  start: 0,
  end: 39,
  loc: 
   SourceLocation {
     start: Position { line: 1, column: 0 },
     end: Position { line: 1, column: 39 },
     lines: Lines { length: 1, name: null, [Symbol(recastLinesSecret)]: [Object] },
     indent: 0 },
  specifiers: 
   [ Node {
       type: 'ImportNamespaceSpecifier',
       start: 7,
       end: 26,
       loc: [Object],
       local: [Function],
       id: null,
       name: null } ],
  importKind: 'value',
  source: 
   Node {
     type: 'StringLiteral',
     start: 32,
     end: 39,
     loc: SourceLocation { start: [Object], end: [Object], lines: [Object], indent: 0 },
     extra: { rawValue: 'react', raw: '\'react\'' },
     value: 'react',
     regex: null } }
```
So it seems like if it would strip out all of the metadata like `start`, `end`, `loc`, and `extra`, then it would be nearly ideal for specifying searches!

ElonVolo:
This is a neat idea, but there hasn't been any action on it in four years and the previous maintainers have moved onto other things. Closing for me.

Issue #165, opened 2016-10-22
Template params swap in class definition


robz:
This:
```
const a = j.identifier('a');
const b = j.identifier('b');
    
const statement = j.template.statement`
  class X extends ${a} {
    f(${b}) {
    }
  }
`;
```
Will output this:
```
class X extends b {
  f(a) {
  }
}
```

The param name and parent class name are swapped.

http://astexplorer.net/#/sftI9tr4Sn

fkling:
Huh, strange. Thanks for reporting!

PR #166, opened 2016-10-22
Handle template replacement when recast traverses out of order

robz:
This fixes issue #165


fkling:
Thank you, this is great!

Issue #148, opened 2016-09-14
Multiple transform support


tivac:
I've got a [collection of transforms](https://github.com/tivac/mithril-codemods) that don't really make sense to be run individually. I'd like to provide a straightforward way to run them all against a group of files. I don't see a way to do this with `jscodeshift` currently, but it seems useful.

[`ava-codemods`](https://github.com/avajs/ava-codemods/blob/f8637ec80c25be2032421fdea2ab5dcd363ab7df/cli.js#L23-L29) does this by running `jscodeshift` once per transform. It's a bit brute-force, but it'll work for now until (if?) `jscodeshift` adds support.

fkling:
That seems very reasonable to me. As a first step, we can pass multiple transforms to the workers and have them apply each one in series (will have to check how custom parser selections impacts this, but I think it should work without problem).

Later we could avoid code generation from the AST and provide a way to "chain" transforms and reuse the existing AST. That would require some thought around custom parsers though.

nickmccurdy:
We could use package based collections of transforms and modify the jscodeshift CLI to support running all of a package's transforms in sequence.

Cross post: https://github.com/facebook/jscodeshift/issues/126#issuecomment-265649259

benjamn:
Recast doesn't care what you do to the AST between parsing and printing, so there shouldn't be any trouble chaining transforms there. You might want to reuse the same `NodePath` tree, though, so you don't have to recreate those objects for each transform.

stevecooperorg:
A cheap and cheerful approach is to create a transform which uses the output of one transform as the input to the next. This is working fairly well for me; wouldn't call it production code but for the hacking I've been doing it suffices. Might be useful to someone else;

```
import fix1 from './fix1'
import fix2 from './fix2'
import fix3 from './fix3'

module.exports = function(file, api, options) {
    const fixes = [fix1,fix2,fix3];
    let src = file.source;
    fixes.forEach(fix => {
        if (typeof(src) === "undefined") { return; }
        src = fix({ ...file, source:src }, api, options);
    });
    return src;
};

```

stkao05:
@stevecooperorg Thanks for sharing the script 👍 Here is a slightly modified version which handles the scenarios where some transform function could return nothing. 

```
import fix1 from './fix1'
import fix2 from './fix2'
import fix3 from './fix3'

module.exports = function(file, api, options) {
    const fixes = [fix1,fix2,fix3];
    let src = file.source;
[...]

PR #171, opened 2016-12-09
Multiple transform support for Workers

### LINKED ISSUE #126: Request: Support codemods published to npm

nickmccurdy:
Fixes #126 and #148.
- [x] Allow the `--transform` flag to be repeated to run multiple transforms in order.
- [x] Change the Runner's API to support multiple transforms. ~~I wanted to do this at first, but I wasn't sure if it would make sense to add a potentially breaking change to this PR.~~
- [x] Support passing a directory or npm package instead of specific files.
- [ ] Keep the AST in memory between each transform, instead of writing and reading to disk between transforms.
- [ ] Test all new use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Usage: jscodeshift <path>... [options]
 path     Files or directory to transform
 
 Options:
-   -t FILE, --transform FILE   Path to a transform file. Can be either a local path or url. Can be given multiple times to run several transforms in order  [./transform.js]
+   -t NAME, --transform NAME   Path to a transform url, file, directory, or npm package. Can be given multiple times to run several transforms in order  [.]
    -c, --cpus                  (all by default) Determines the number of processes started.
    -v, --verbose               Show more information about the transform process  [0]
    -d, --dry                   Dry run (no changes are made to files)
@@ -47,8 +47,8 @@ Options:
 ```
 
 This passes the source of all passed through the transform modules specified
-with `-t` or `--transform` (they default to a `transform.js` file in the current
-directory). The next section explains the structure of transform modules.
+with `-t` or `--transform` (they default to the current directory). The next
+section explains the structure of transform modules.
 
 ## Transform module
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Usage: jscodeshift <path>... [options]
 path     Files or directory to transform
 
 Options:
-   -t FILE, --transform FILE   Path to the transform file. Can be either a local path or url  [./transform.js]
+   -t FILE, --transform FILE   Path to a transform file. Can be either a local path or url. Can be given multiple times to run several transforms in order  [./transform.js]
    -c, --cpus                  (all by default) Determines the number of processes started.
    -v, --verbose               Show more information about the transform process  [0]
    -d, --dry                   Dry run (no changes are made to files)
@@ -46,13 +46,13 @@ Options:
    --version                   print version and exit
 ```
 
-This passes the source of all passed through the transform module specified
-with `-t` or `--transform` (defaults to `transform.js` in the current
-directory). The next section explains the structure of the transform module.
+This passes the source of all passed through the transform modules specified
+with `-t` or `--transform` (they default to a `transform.js` file in the current
+directory). The next section explains the structure of transform modules.
 
 ## Transform module
 
-The transform is simply a module that exports a function of the form:
+A transform is simply a module that exports a function of the form:
 
 ```js
 module.exports = function(fileInfo, api, options) {
--- a/README.md
+++ b/README.md
@@ -48,7 +48,20 @@ Options:
 
 This passes the source of all passed through the transform modules specified
 with `-t` or `--transform` (they default to the current directory). The next
-section explains the structure of transform modules.
+section explains how transform modules are located, and the section after that
+explains the structure of transform modules.
+
+## Transform location algorithm
+The default transform is the current directory if no `--transform` flag is given.
+
+1. If URL, return file at URL
+2. If file, return file
+3. If directory
+  1. If contains directory named `transforms`, return files in that directory
+  2. If contains file named `transform.js`, return file
+  3. Else, return files in directory
+4. If npm package, find its directory (with `require.resolve()`) and go to step 3
+5. Else, error
 
 ## Transform module
 
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -18,7 +18,7 @@ const fs = require('fs');
 const path = require('path');
 const http = require('http');
 const https = require('https');
-const temp = require('temp');
+const temp = require('temp').track();
 const ignores = require('./ignoreFiles');
 
 const availableCpus = Math.max(require('os').cpus().length - 1, 1);
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -135,7 +135,7 @@ function getAllFiles(paths, filter) {
   ).then(concatAll);
 }
 
-function run(transformFile, paths, options) {
+function run(transformFiles, paths, options) {
   let usedRemoteScript = false;
   const cpus = options.cpus ? Math.min(availableCpus, options.cpus) : availableCpus;
   const extensions =
@@ -147,6 +147,9 @@ function run(transformFile, paths, options) {
   ignores.add(options.ignorePattern);
   ignores.addFromFile(options.ignoreConfig);
 
+  transformFiles = Array.isArray(transformFiles) ? transformFiles : [transformfiles]
+
+  let transformFilePromises = transformFiles.map(transformFile => {
   if (/^http/.test(transformFile)) {
     usedRemoteScript = true;
     return new Promise((resolve, reject) => {
@@ -163,7 +166,7 @@ function run(transformFile, paths, options) {
               fs.write(info.fd, contents);
               fs.close(info.fd, function(err) {
                 reject(err);
-                transform(info.path).then(resolve, reject);
+                  resolve(info.path);
               });
             });
         })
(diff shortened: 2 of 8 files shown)


fkling:
Thank you for starting to work on this! Let me comment on your ideas:

> Change the Runner's API to support multiple transforms. I wanted to do this at first, but I wasn't sure if it would make sense to add a potentially breaking change to this PR.

I actually think this is what we should do and it doesn't necessarily require a breaking change. The runner could accept a single transform (which is a string anyway) or an array of transforms and pass them to the worker.

As it is now, there probably wouldn't be any performance gains compared to invoking the CLI repeatedly. That said, there may not be any performance gains from doing this in worker instead either, however we could avoid writing intermediate results to disk.

I'd be curious to see if there are any performance difference between your current implementation and the one I mentioned.

> Keep the AST in memory between each transform, instead of writing and reading to disk between transforms.

This is something we should definitely do, eventually. I believe this requires a more general discussion about the contract between transformers and jscodeshift (i.e. what will the structure of the return value of the transformer be), so no rush here.

> Support passing a directory of transforms, a directory with a transforms directory, or an npm package instead of specific files.

I also think this requires a discussion about how transforms published to the npm should be structured.

nickmccurdy:
> The runner could accept a single transform (which is a string anyway) or an array of transforms and pass them to the worker.

I changed my mind and started implementing this. Right now the Runner and Workers can each take multiple transforms, and a single transform is converted to a list of only that transform for API compatibility. I'm also using JSON to serialize transform lists when spawning Workers in child processes. 

>> Keep the AST in memory between each transform, instead of writing and reading to disk between transforms.

> This is something we should definitely do, eventually.

This would be great, and I remember seeing you comment about this in another issue. This is my first time digging into the codebase, so I'm still trying to get a sense of how easily this could be done without changing the public API significantly.

> I also think this [a convention for transform directories/packages] requires a discussion about how transforms published to the npm should be structured.

As far as I'm aware, most popular codemod packages store transform modules directly in a top-level `transforms` directory, and some of them are also published on npm. Normally you'd manually find their path in `node_modules` and get the right subfolder, but I was thinking that the path lookup could be automated by giving just the package name and letting jscodeshift's CLI handle path lookup and listing transforms. If you're already in a `transforms` directory or a parent of one, it could run those transforms by default as a convenience (if you give no `-t` option). As far as I'm aware, these features shouldn't require API breaking changes for transform authors.

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
-    process.stderr.write(
-      colors.white.bgRed('ERROR') + ' Transform file ' + transformFile + ' does not exist \n'
-    );
-    return;
-  } else {
-    return transform(transformFile);
+  } else if (fs.existsSync(transform)) {
+    let stats = fs.lstatSync(transform);
+
+    if (stats && stats.isFile(transform)) {
```
fkling: I believe `lstatSync` always returns something, so testing for the truthiness of `stats` is unnecessary. [`isFile` doesn't accept an argument](https://nodejs.org/api/fs.html#fs_class_fs_stats).

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+      // handle file
+      return transform;
+    } else if (stats && stats.isDirectory(transform)) {
```
fkling: Same as above.

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+      let directory = path.join(transform, 'transforms');
+      let file = path.join(transform, 'transformjs');
```
fkling: Typo here: `transformjs` -> `transform.js`

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+
+      // handle directory
+      if (fs.existsSync(directory) && fs.lstatSync(directory).isDirectory()) {
+        return fs.readdirSync(directory);
```
fkling: I feel like it might be worth to to be a more defensive here and filter out anything that is not a JS file (at least anything that is a directory).

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+      } else if (fs.existsSync(file) && fs.lstatSync(file).isFile()) {
+        return file;
+      } else {
+        return fs.readdirSync(transform);
```
fkling: Same as above.

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+      }
+    } else {
+      // handle other fs objects
+      return transform;
```
fkling: Maybe it would be better to throw an error here? In which cases would this code path be hit?

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+    }
+  }
+
+  try {
+    // handle npm package as directory
+    return getTransform(require.resolve(transform));
```
fkling: I don't think that works as planned. Or at least it works differently than you mentioned in the README. `require.resolve(...)` will either resolve to a file path or it will throw an error if the module doesn't have an `index.js` file and the `package.json`'s `main` field is not set.

That means that with the current logic it would go back to step 2, not 3, in the README.

However, I feel like it could be valuable to be able to publish package with multiple transforms. Do you expect users to to `-t someTransform/transforms/doA.js -t someTransform/transforms/doB.js` ?

If you want `-t someTransforms` to work, then some additional work needs to be done to get the package root. [find-root](https://www.npmjs.com/package/find-root) could help with that.

src/Runner.js
```diff
@@ -172,16 +161,63 @@ function run(transformFile, paths, options) {
...
+    options.extensions && options.extensions.split(',').map(ext => '.' + ext);
+  const fileCounters = {error: 0, ok: 0, nochange: 0, skip: 0};
+  const statsCounter = {};
+  const startTime = process.hrtime();
 
-  function transform(transformFile) {
+  ignores.add(options.ignorePattern);
+  ignores.addFromFile(options.ignoreConfig);
+
+  transforms = Array.isArray(transforms) ? transforms : [transformfiles]
```
fkling: I believe this should be `[transforms]` here?

README.md
```diff
@@ -46,13 +46,26 @@ Options:
...
+## Transform location algorithm
+The default transform is the current directory if no `--transform` flag is given.
+
+1. If URL, return file at URL
+2. If file, return file
+3. If directory
+  1. If contains directory named `transforms`, return files in that directory
+  2. If contains file named `transform.js`, return file
+  3. Else, return files in directory
+4. If npm package, find its directory (with `require.resolve()`) and go to step 3
```
fkling: I think it would be good to point out here that the package name is resolved relative to where jscodeshift is installed. So if jscodeshift is installed globally, it would look for a globally installed package. If it's installed locally, it looks for a locally installed package.

Or am I getting this wrong?

src/Runner.js
nickmccurdy: My mistake, I should be testing to ensure that `stats` isn't an `Error`.

src/Runner.js
nickmccurdy: I was originally planning on filtering directories, but I must have forgotten. If we do filter out non-JS files, do we need to support multiple extensions (not `.js`) to support transpilers?

README.md
nickmccurdy: I believe it's based off the module paths (which `require()` and `require.resolve()` use internally), so that should be right. I assumed this would be obvious enough to users since module paths for local packages are separated from global packages. There are other Node tools (including Karma and Grunt) that rely on plugin modules also being installed locally/globally. However, I could add a note about this if you don't think it would be obvious.

fkling: I could imagine that people might think jscodeshift is doing something magical to look in the current folder for a matching package. I think it doesn't hurt to clarify what happens.

src/Runner.js
fkling: I think it isn't too much to ask that transform files always end in `.js`.

nickmccurdy:
@fkling Sorry for the delay, I got stuck on some testing issues and was distracted. The main thing I have left is testing `getTransform()` with different directory structures to ensure that all of the use cases I added to the readme are unit tested.

However, I ran into some roadblocks because of #173 (basically, I need to frequently setup and teardown a folder to use as `getTransform()` subject, but the current tempfile code in the test suite does not delete temp files between tests, causing tested use cases to fail because there are too many files in the same directory). I started a pull request to fix #173 but I was distracted, I'll take another look soon and hopefully I can iron out the bugs I was running into. I might ask you for some advice if you don't mind.

ElonVolo:
Closing for now since there hasn't been any activity on this. I would, however, like to compliment the participants on the hard work that went into this PR, and hopefully we can re-address this feature in the future.

PR #174, opened 2016-12-11
Simplify npm scripts (for Travis and dev)

nickmccurdy:
## Remove `npm test` from `npm run prepublish`
I noticed that Travis runs jest twice with the same results on every build. This happens because Travis runs `npm install` and `npm test` by default, but this package's prepublish hook (which is executed by `npm install`) also runs `npm test`. If you aren't seeing this, make sure you're expanding the `npm test` command in the Travis build log.

To get the same behavior as `npm install` before, run `npm install-test`.

## Don't copy files to `dist`
It seems like the `cp -R src/ dist/` command was added to support the same directory structure while removing a preprocessor, but I don't think it's needed anymore. I was confused by `npm test` not using the latest versions of my source files, so I thought this may help.


fkling:
Good call with the prepublish command. However, I'm hesitant to remove the `dist/` directory. You are right, it's not needed anymore (internally), but I left it in to avoid any breaking API changes since there might be clients that directly reach into `dist/`.

nickmccurdy:
Hm, is the dist directory documented? Worst case it could be be a major/minor breaking change, or the src directory could be renamed to dist.

fkling:
It's not documented, but a `dist/` folder seems to be imply that it's fine to access it since it contains build artifacts. At least that's how I interpret it.

But I understand that it's weird for testing. As a compromise, how about `bin/Runner.js` requires modules from `src/` instead of `dist/` and we remove `/src/` from `.npmignore`?

nickmccurdy:
Good point, could we keep all the changed references (from `dist` to `src`) while replacing the `prepublish` script with just `cp -R src/ dist/` and using your `.npmignore` suggestion? If I'm understanding this correctly, this would mean that we could completely ignore `dist` for testing without having to write to it, while including it in npm builds as a duplicate of `src` to avoid breaking changes.

fkling:
Yes, that's exactly what I mean 👍

nickmccurdy:
Cool, how does this look now? Want me to squash against master?

nickmccurdy:
This Travis build has been stalling for a while, any ideas? It took under 12 minutes on my fork.

fkling:
Sometimes there are no resources for the CI job to run. I restarted the build and it passed.

README.md
```diff
@@ -345,7 +345,7 @@ To define a test, use `defineTest` from the `testUtils` module:
 
 ```js
 jest.autoMockOff();
-const defineTest = require('jscodeshift/src/testUtils').defineTest;
+const defineTest = require('jscodeshift/dist/testUtils').defineTest;
```
nickmccurdy: Since src and dist are the same files (assuming `prepublish` was recently run), what are the advantages of recommending dist over src here?

fkling: Just a precaution.  I don't want anybody relying on `src/` having a certain structure or contain certain files. `dist/` is the folder that is "exposed".

nickmccurdy:
Thanks for your help. 🎉

Issue #44, opened 2015-08-23
File comment is lost when removing first statement


spicyj:
If you have

```
/**
 * @providesModule Foo.react
 */

var React = require('React');
```

and remove the require statement, the docblock is lost. This causes build errors and I had to manually fix up a bunch of files with the last codemod I wrote to restore these.

fkling:
Mmh. I assume the comment is attached to the VarableDeclaration, so when you remove the node, the comment is removed as well. You could save the comment first and reattach it to the file node: http://felix-kling.de/esprima_ast_explorer/#/kGspd21ZnD/1.

Maybe we can do something to make this easier, but not sure yet what.

spicyj:
Yeah, I figured as much. Maybe we could just special-case the first comment? It seems rare that you'd ever want to move or drop it (especially at FB).

udnisap:
I can work on it to detach the first comment a file from the first statement and add as a node. Is this desirable?

DrewML:
@fkling What are your thoughts on adding a `recipes` section to the documentation (similar to [this](https://github.com/avajs/ava/tree/master/docs/recipes)). I've found a few examples of use-cases in the past digging through the issues on this repo that would have been great to have documented.

I know there are already links to the codemod repos in the bottom of the readme, but specific recipes could be an added bonus.

fkling:
@DrewML: Sounds like an excellent idea to me!

slorber:
Hey,

By chance could someone take a look at this issue?

https://github.com/5to6/5to6-codemod/issues/18


The problem here for me is that sometimes the codemod does try to restore the leading comment, but that leading comment was not removed in the first place, leading to a duplicate leading comment.

I'm not a JSCodeShift/Recast expert but couldn't we make sure the actual leading comment gets removed before restoring it, or restoring it only when necessary?

ptbrowne:
@slorber Yes it seems that if the first node was not modified, the first comment would be duplicated.

The solution is to check if the first node has changed : http://astexplorer.net/#/hb6iLO9hTe/48

danielo515:
What is the rationale to attach comments to other nodes? Why can't they live as independent nodes?

PR #181, opened 2017-01-11
Update first-comment recipe for when the first node is not updated

ptbrowne:
The comments on the first line would be duplicated if the first node was not modified.

https://github.com/5to6/5to6-codemod/issues/18
https://github.com/facebook/jscodeshift/issues/44#issuecomment-271335348


fkling:
This looks great, thank you!

PR #187, opened 2017-03-14
Methods to find JSX elements from ES6 imports

chrislloyd:
This PR adds two extra methods to Collection that helps people find JSXElements when using ES6 imports. An example:

```jsx
import Foo from 'foo';
import { Bar as NewBar } from 'bar';

<Foo />
<NewBar />
```

```jsx
nodes.findJSXElementsByImport('foo'); // returns <Foo />
nodes.findJSXElementsByNamedImport('bar', 'Bar'); // returns <NewBar />
```

Tests included.

### DIFF
--- a/src/collections/JSXElement.js
+++ b/src/collections/JSXElement.js
@@ -65,7 +65,64 @@ const globalMethods = {
             .paths();
         }
       });
-  }
+  },
+
+  /**
+   * Finds all JSXElements by package name. Given
+   *
+   *     import Bar from 'Foo';
+   *     <Bar />
+   *
+   * findJSXElementsByImport('Foo') will find <Bar />, without having to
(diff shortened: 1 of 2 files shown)


TheSavior:
Why not make the existing `findJSXElementsByModuleName` support the import syntax? The named one is a bit more unique.

vamshi9666:
any updates on this ? @TheSavior.

ElonVolo:
I see now. There’s already an existing “find jsx by require()” functionality, and if we’re going to keep consistency, we need to include support of ES6/Typescript imports.

I’m not sure where this fits in with jscodeshift’s “no new features” mandate, but if it doesn’t, I can add the functionality to the evcodeshift fork.

Issue #47, opened 2015-09-02
Programmatic Usage


thejameskyle:
The CLI works great, but it'd be nice to have a friendly API for using jscodeshift within a node script (I'm attempting to use it within yeoman).

`Runner.run` works, but there's no callback, stream, or promise returned so you can't tell when it's finished. Maybe that's all that needs to be changed.

cpojer:
What kind of programmatic usage are you looking for? You can `require('jscodeshift')` and use its API like you would in a transform. We haven't made an effort to split out the runner from jscodeshift but you can always use it by calling exec.

thejameskyle:
Yeah, I was trying to do this:

```js
var jscodeshift = require('jscodeshift');
jscodeshift.run(...);
```

However, `run` does stuff asynchronously without returning the promise here: https://github.com/facebook/jscodeshift/blob/master/src/Runner.js#L96

PR #190, opened 2017-03-24
expose Runner and get access to transform results

ganarajpr:
Access to the runner from the jscodeshift api - so that you can programmatically run jscodeshift ( same as running it from cli ). Currently I am not sure how you can do this without this PR. 

This seems to have been requested before ( #47 ) and perhaps was implemented but seems to have been missed recently ?

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -119,6 +119,15 @@ class Collection {
   }
 
   /**
+   * Provides a callback mechanism to be notified when done with previous steps.
+   *
+   * @param {function} callback
+   */
+  done(cb) {
+    return cb.call(this);
+  }
+
+  /**
    * Returns the number of elements in this collection.
    *
    * @return {number}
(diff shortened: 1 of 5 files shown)


joyarzun:
Hi there
Can you please review this PR? For now it's the best option to call Jscodeshift without cli.
Thanks

ganarajpr:
@fkling Any update on this?

src/Collection.js
```diff
@@ -118,6 +118,15 @@ class Collection {
     return this.__paths.length;
   }
 
+  /**
+   * Provides a callback mechanism to be notified when done with previous steps.
+   *
+   * @param {function} callback
+   */
+  done(cb) {
```
fkling: Which use case does this solve? Why not do

```
Collection.fromNodes(...).forEach(...);
// do stuff afterwards
```

The `Collection` API is not asynchronous.

buoyantair:
@ganarajpr Are you still interested in working on this? It would be really nice to have some way to run transformations & get status etc.

PR #196, opened 2017-05-14
template.expression should not parse string without interpolation as statement

jsnajdr:
Let's parse a function expression with `jscodeshift.template.expression`:
```js
expression`function() {}`
```

This code is a valid expression, but not a valid statement -- a function needs to have a name to be a valid statement.

`template.expression` surrounds the code with `()` to force parsing as expression. However, it fails to do so if the template string doesn't have any interpolation parameters. Therefore, the above function expression fails to parse.

This PR fixes that bug -- the length of the `template` array is incorrectly checked. It should check for `length === 1` instead of `length === 0`. Moreover, a special case for `length === 1` is not even necessary.

### DIFF
--- a/src/__tests__/template-test.js
+++ b/src/__tests__/template-test.js
@@ -111,6 +111,18 @@ while (i < 10) {
     .toEqual(expected);
   });
 
+  it('correctly parses expressions without any interpolation', () => {
+    const expected = 'function() {}';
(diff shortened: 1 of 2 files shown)


xixixao:
This indeed fixes a bug and looks good to merge.

Issue #197, opened 2017-05-25
Update Babel-Core to avoid minimatch issues on CI builds


ev1stensberg:
I get this error:
![skjermbilde 2017-05-25 kl 12 38 20](https://cloud.githubusercontent.com/assets/16735925/26447060/10ad6a70-4147-11e7-842a-930003093de5.png)
Along the CI build at:

https://circleci.com/gh/webpack/webpack/840?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Any way you could bump dependencies?

fkling:
See my explanation at https://github.com/facebook/jscodeshift/pull/198#issuecomment-304329657

PR #198, opened 2017-05-25
Bump Babel Core dep in accordance to minimatch issue

ev1stensberg:
#197 Related. Close if unwanted.


fkling:
Yeah, this is not something we can do because we need babel v5 for backwards compatibility. Maybe if we release a new breaking version we could remove babel v5 support, but not at the moment.

Thank you though!

Issue #203, opened 2017-06-15
Test Util for defining input and output inline


TheSavior:
I want to write tests for my codemod. My codemod messes with the require statements at the top of the file so I have to have totally separate input fixtures. The current testUtil `defineTest` requires an individual file for each input/output pair. 

I'd like to be able to define my input and expected output inline, passing a reference to the transform, like this:

```
const defineInlineTest = require('jscodeshift/dist/testUtils').defineInlineTest;
const myTransform = require('../myTransform');

defineInlineTest(myTransform, 
`
  const missing = missingRequire('foo');
`, 
`
  const missing = require('foo');
  const added = require('added');
`);

defineInlineTest(myTransform, 
`
  const added = require('added');
`, 
`
  const added = require('added');
`);
```

It seems like this would also be a step in the right direction to being able to use snapshots as discussed here: https://github.com/facebook/jscodeshift/issues/135.

I'd be happy to take this on.

PR #204, opened 2017-06-16
Adding the ability to write tests inline, without fixture files

TheSavior:
Closes https://github.com/facebook/jscodeshift/issues/203

### DIFF
--- a/src/testUtils.js
+++ b/src/testUtils.js
@@ -15,7 +15,11 @@ const fs = require('fs');
 const path = require('path');
 
 function runInlineTest(module, options, input, expectedOutput) {
+  // Handle ES6 modules using default export for the transform
   const transform = module.default ? module.default : module;
+
+  // Jest resets the module registry after each test, so we need to always get
+  // a fresh copy of jscodeshift on every test run.
   let jscodeshift = require('./core');
   if (module.parser) {
     jscodeshift = jscodeshift.withParser(module.parser);
--- a/src/testUtils.js
+++ b/src/testUtils.js
@@ -14,6 +14,25 @@
 const fs = require('fs');
 const path = require('path');
 
+function runInlineTest(module, options, input, expectedOutput) {
+  const transform = module.default ? module.default : module;
+  let jscodeshift = require('./core');
+  if (module.parser) {
+    jscodeshift = jscodeshift.withParser(module.parser);
+  }
+
+  const output = transform(
+    input,
+    {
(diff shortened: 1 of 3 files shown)


src/testUtils.js
```diff
@@ -47,25 +66,10 @@ function runTest(dirName, transformName, options, testFilePrefix) {
...
-
-  // Jest resets the module registry after each test, so we need to always get
-  // a fresh copy of jscodeshift on every test run.
-  let jscodeshift = require('./core');
-  if (module.parser) {
-    jscodeshift = jscodeshift.withParser(module.parser);
-  }
-
-  const output = transform(
-    {path: inputPath, source},
```
TheSavior: The one thing that feels a bit gross in this change is that inline tests aren't able to have a path here. It seems like most transforms only use source, but people will be unable to use inline tests with their transform if they use the file path directly.

fkling:
cc @Daniel15 who built the test utils

Daniel15:
This seems reasonable to me... I'll take a closer look when I'm at my
computer.

I've also been meaning to change the test util to use Jest's snapshot
feature (keep the existing stuff for backwards compatibility, and use
snapshots when an output file isn't found so it's used for any future
tests).

TheSavior:
Yeah, I was thinking about the snapshot test option as I was writing this. I think the snapshot test piece will be able to use the `runInlineTest` function in this pr under the hood.

src/testUtils.js
```diff
@@ -47,25 +66,10 @@ function runTest(dirName, transformName, options, testFilePrefix) {
   );
   // Assumes transform is one level up from __tests__ directory
   const module = require(path.join(dirName, '..', transformName + '.js'));
-  // Handle ES6 modules using default export for the transform
-  const transform = module.default ? module.default : module;
-
-  // Jest resets the module registry after each test, so we need to always get
```
Daniel15: It looks like these comments have gone missing - Could you please re-insert them?

Daniel15:
This looks pretty good to me! Please re-insert the comments that have been deleted:
```
// Handle ES6 modules using default export for the transform
```
```
// Jest resets the module registry after each test, so we need to always get
// a fresh copy of jscodeshift on every test run.
```

And I'll merge this 😃

TheSavior:
Whoops. Good catch. Updated.

Daniel15:
Thank you! :D

PR #208, opened 2017-06-22
Add `dynamicImport` plugin to `babylon` parser.

umidbekkarimov:
[Dynamic import](https://webpack.js.org/guides/code-splitting-async/#dynamic-import-import-) is widely used in webpack apps and it would be great if `jscodeshift` start to support it by default.


kentcdodds:
Ping. How can I help get this merged?

PR #211, opened 2017-07-17
Fix `renameTo` not taking property shorthands into account

vitorbal:
This fixes the `renameTo` helper method not properly renaming object properties that are shorthands.

Bug reproduction in AST Explorer: http://astexplorer.net/#/gist/0ad1434f08e1720089d7085fbdfe7ba0/6f6632f0bb6e2e3aa77d5e873d0ab05d6c3be0e3

The previous version of the code would rename the property's `value` correctly, but since `shorthand` was still set as `true`, the output result would still be the old variable as a shorthand
property.

Now, if an identifier to be renamed is of type `Property` and is a shorthand (but not a method), we flip the `shorthand` value to be `false`.

Added a test for this as well.

### DIFF
--- a/src/collections/VariableDeclarator.js
+++ b/src/collections/VariableDeclarator.js
@@ -126,6 +126,17 @@ const transformMethods = {
             scope = scope.parent;
           }
           if (scope) { // identifier must refer to declared variable
+
+            const parent = path.parent.node;
+            if (
+              types.Property.check(parent) &&
+              parent.shorthand &&
+              !parent.method
+            )  {
+
+              path.parent.get('shorthand').replace(false);
+            }
+
             path.get('name').replace(newName);
           }
         });
--- a/src/collections/VariableDeclarator.js
+++ b/src/collections/VariableDeclarator.js
@@ -134,7 +134,11 @@ const transformMethods = {
             scope = scope.parent;
           }
           if (scope) { // identifier must refer to declared variable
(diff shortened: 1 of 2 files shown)


vitorbal:
Hmm that failure looks unrelated to my changes 🤔 

```
 FAIL  src/__tests__/Worker-test.js (17.716s)
  ● Worker API › custom parser › uses babylon if configured as such
    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
      
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:523:19)
      at ontimeout (timers.js:488:11)
      at tryOnTimeout (timers.js:323:5)
      at Timer.listOnTimeout (timers.js:283:5)
```

baopham:
🤔 what if I want to rename it like this though:

```js
// Outputs:
var newFoo = 42;
var obj2 = {
  newFoo,
};
```

instead of:

```js
// Outputs:
var newFoo = 42;
var obj2 = {
  foo: newFoo,
};
```

vitorbal:
@baopham I think that would be out of scope for this bug fix since you'd have to find all the places in the code that depend on `obj2.foo`.

baopham:
Yeah, agree that it's out of scope.

PR #216, opened 2017-08-18
Add some() and every() methods for Collection

baopham:


baopham:
Travis seems stuck...

PR #218, opened 2017-08-19
Allow specifying a testName for inline tests

iancw:
This is a small change that makes test output a bit nicer when writing
multiple inline tests.


fkling:
Seems reasonable, thanks!

PR #227, opened 2017-11-01
Bump recast to v0.12.8

0x24a537r9:
I'm primarily interested in fixing the issues around type spreading and opaque types from:
  - https://github.com/benjamn/recast/pull/431
  - https://github.com/benjamn/ast-types/pull/232
  - https://github.com/benjamn/recast/pull/438

Test plan: `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "micromatch": "^2.3.7",
     "node-dir": "0.1.8",
     "nomnom": "^1.8.1",
-    "recast": "^0.12.5",
+    "recast": "^0.12.8",
     "temp": "^0.8.1",
     "write-file-atomic": "^1.2.0"
   },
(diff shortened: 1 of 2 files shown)


arahansen:
I'm also running into this issue with type spreading. Would love to see this merged!

0x24a537r9:
@fkling Could I get this merged?

0x24a537r9:
Ah, nvm. I misunderstood the caret range syntax--`^0.12.5` will already match anything `0.12.X` according to the [caret range docs](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004). As such, there's no reason for this change.

@arahansen You should be able to just rerun yarn or npm and you'll get the latest now.

benjamn:
Just a friendly note that `recast@0.13.0` is out now. The biggest change is that `core-js` is no longer a dependency, since I've dropped support for Node <4.

Issue #229, opened 2017-11-10
Update/Remove babel-core in a new major version


sokra:
Currently jscodeshift depends on an old version of babel-core, which depends on `minimatch@2` which
* Reports a security warning on `npm install` pointing to a security issue
* Reports errors for `nsp` which checks for security warnings

Because of these two problems I'm unwilling to use this package. These warnings/errors are transitive and I don't want my users to see security issues reported. Note that this is not a real security issue for jscodeshift, because it's not using the affected babel options. Nevertheless I don't want to tell every user to ignore these warnings, because it affects the trust in security warnings!

In another issue you pointed out that removing/updating babel would be a backward incompatible change:

see https://github.com/facebook/jscodeshift/pull/198#issuecomment-304329657

> Yeah, this is not something we can do because we need babel v5 for backwards compatibility. Maybe if we release a new breaking version we could remove babel v5 support, but not at the moment.

So I propose to increase the major version and release a breaking change.

Do you really want a security warning reported on every npm install of your package? This affects the trust in security warnings! This trust is important for the community! Especially a package under the facebook namespace should be an example.

slorber:
@sokra as mentionned on Twitter, I was able to run successfully jscodeshift on babel6 by backporting the `babel.parse` function from babel5 to babel6. 

This function has been deleted in babel6 and in babel5 it was just a wrapper around Babylon parser with some preset.

Check source code here for backport code:
https://github.com/slorber/react-reboot/blob/master/src/lib/react-reboot.js#L23

TheLarkInn:
@fkling Is there anything we can help or do to unlock this? Happy to help in any way. This is a top 5 priority item for us to unlock merging webpack-cli into webpack which is a key integration for webpack 4. 

cc @cpojer incase notification aren't making it to Felix (which they do with me most times).

ematipico:
@slorber I don't think it's a matter of make it working (thank you very much for the workaround provided), but it's more like the feedback that the library is going to provide during the installation. 

How would you feel, as end-user, to see that an OSS library from an important company shows a warning related to a security issue? I thinks that's the message of the argument. Security patches have to provided.

slorber:
@sokra @ematipico to me it is not really clear, is one of your libraries importing jscodeshift? or do you worry about direct users of jscodeshift?

@ematipico the thing is, if we can make it work with babel6, then you can upgrade safely and won't get the warning again.

[...]

PR #230, opened 2017-11-24
Replace babel-5 parser with retrocompatible babylon setup

slorber:
See https://github.com/facebook/jscodeshift/issues/229

We can get rid of Babel5 as the dependency on Babylon is enough to make the tests pass, as long as we use Babylon in the same way Babel5 used to use it under the hood


sokra:
@fkling Could you take a look at this? I would really great to get rid of the security warning...

PR #238, opened 2018-02-13
Guarantee uniqueness of variable names

somewhatabstract:
The current code causes duplicate variable names due to there being no guarantee that `Math.random()` will return non-colliding values. This is especially true given the 32-bits of the V8 `Math.random()` implementation.

This change replaces the use of `Math.random()` with a simple counter to guarantee that each name returned is different.

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -103,8 +103,9 @@ function replaceNodes(src, varNames, nodes, parser) {
   return ast;
 }
 
-function getRandomVarName() {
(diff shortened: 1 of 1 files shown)


fkling:
cc @TheSavior . Maybe this solves the problem you were seeing in unit tests.

TheSavior:
Ooh ooh. Yeah, looks likely.

PR #239, opened 2018-02-22
Bump recast and babylon to support JSX fragments

xixixao:
What is says on the can.


xsburg:
Just wanted to point out that this PR basically adds support for TypeScript transformations as well, thanks to the effort of @benjamn and @Phoenixmatrix who supported it in the `ast-types` and `recast` packages.
All we need now is to create a babylon parser with enabled `typescript` option:
```javascript
module.exports = function(fileInfo, api) {
  // your codemod here
}
const babylon = require('babylon');
const parserOptions = {
    sourceType: 'module',
    allowImportExportEverywhere: true,
    allowReturnOutsideFunction: true,
    plugins: [
        'jsx',
        'typescript',
        'asyncFunctions',
        'classConstructorCall',
        'doExpressions',
        'trailingFunctionCommas',
        'objectRestSpread',
        'decorators',
        'classProperties',
        'exportExtensions',
        'exponentiationOperator',
        'asyncGenerators',
        'functionBind',
        'functionSent',
        'dynamicImport'
    ]
};
module.exports.parser = {
    parse(code) {
        try {
            return babylon.parse(code, parserOptions);
        } catch (e) {
            debugger;
        }
    }
};
```

benjamn:
With `recast@0.14.2` (just published), @xsburg's example becomes even easier:
```js
module.exports = function(fileInfo, api) {
  // your codemod here
}
// Use a parser (Babylon 7) preconfigured for TypeScript.
// Note that you still need to run `npm install babylon@next` yourself,
// since Recast does not depend directly on Babylon.
module.exports.parser = require("recast/parsers/typescript");
```
[More information](https://github.com/benjamn/recast#using-a-different-parser) about passing a different parser to `recast.parse`. The preconfigured parsers are still a work in progress, so please let me know if they fail to meet your needs.

Phoenixmatrix:
I believe there's still gonna be the issue of jsx vs type assertions (not using the "as" syntax) being mutually exclusive though?

benjamn:
Yeah, `recast/parsers/typescript` actually doesn't include the `jsx` Babylon plugin.

PR #241, opened 2018-03-04
Perf: replace async with neo-async

dfederm:
neo-async is a drop-in replacement for async which is faster.


fkling:
Seems fine to me :)

PR #242, opened 2018-03-05
Fix error handling for http

BridgeAR:
The reject handler were always called no matter if there was an or not.

Now it will properly test for that and also wait for the write to be done before closing fd.

It will also report an error including a stack trace from now on and not only show a error message in case there is an error in the http(s) call.

I also fixed the issue that the write did not have a callback at all, so that error was always silent so far. This is going to throw from Node.js 10.x on. See https://github.com/nodejs/node/pull/18668.


fkling:
I should reviewed that better... ;) Thanks!

PR #244, opened 2018-03-12
Fix Reactiflux link to dead channel

### LINKED ISSUE #243: Reactiflux doesn't have a #codemod channel
The [Support](https://github.com/facebook/jscodeshift#support) section of the readme seems to link to a Reactiflux channel that no longer exists. Has this moved to another server or should we change the wording to point to Reactiflux's #tooling channel?

nickmccurdy:


mfowlewebs:
Would be stellar if you would add a human-friendly description somewhere in the title.

fkling:
Not sure if people even go to reactiflux, but we should definitely remove the dead channel link. Thank you!

PR #251, opened 2018-04-01
feat(cli): accept code from stdin

sjy:
- accept code from stdin and write transformed result to stdout
- by default apply --run-in-band --silent and --print option as true

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -18,6 +18,7 @@ const http = require('http');
 const https = require('https');
 const temp = require('temp');
 const ignores = require('./ignoreFiles');
+const createTempFileWith = require('../utils/createTempFileWith');
 
 const availableCpus = Math.max(require('os').cpus().length - 1, 1);
 const CHUNK_SIZE = 50;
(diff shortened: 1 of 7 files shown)


stellarhoof:
What's happened to this PR

src/Runner.js
```diff
@@ -48,6 +49,31 @@ function concatAll(arrays) {
...
+        }
+      });
+
+      stdin.on('end', () => {
+        resolve(rst);
+      });
+    }
+  });
+}
+
```
yungsters, line 76: Instead of buffering the entire file from `stdin` and then writing it to disk (only to be re-read again when we process it), can you instead change this to write `process.stdin` directly to an open handle to a temporary file (e.g. using `fs.createWritesStream` and `process.stdin.pipe`)?

src/Runner.js
```diff
@@ -178,14 +204,23 @@ function run(transformFile, paths, options) {
     );
     return;
   } else {
-    return transform(transformFile);
+    return getStdin().then(str => transform(transformFile, str));
```
yungsters, line 207: You should check whether `process.stdin.isTTY` is false and only read from `process.stdin` in that case.

src/Runner.js
```diff
@@ -178,14 +204,23 @@ function run(transformFile, paths, options) {
...
+    )
+      .then(fileList => {
+        let files = fileList;
+        if (strFromStdin) {
+          // only print stdin content to stdout
+          options.runInBand = true;
+          options.print = true;
+          options.silent = true;
+          files.push(createTempFileWith(strFromStdin));
+        }
```
yungsters, line 223: This should happen outside of `transform`.

yungsters (changes requested):
Stepping in to help review pull requests.

Daniel15 (changes requested):
@yungsters' comments

ElonVolo:
There hasn't been any action of this in almost four years. Closing for now, but it's a functionality we could look into again in the future.

PR #265, opened 2018-06-10
PARTIAL FIX - Replace Nomnom with Commander - Looking for feedback

oculus42:
Looking for help on this one. I have replaced `nomnom` with `commander` and the tests pass with the exception of one: `jscodeshift CLI › passes options along to the transform`.

Commander doesn't support arbitrary args, like nomnom, so it can't pass them along.

I did some quick looking and didn't find a good alternative.


oculus42:
haven't made progress on this and no feedback, so I'm closing for now at least.

PR #269, opened 2018-07-18
add a parser option for typscript

matthewrobertson:
Create a new parser option for typescript that uses the typescript plugin for babylon

### DIFF
--- a/bin/jscodeshift.sh
+++ b/bin/jscodeshift.sh
@@ -85,10 +85,10 @@ const opts = require('nomnom')
       help: 'No output'
     },
     parser: {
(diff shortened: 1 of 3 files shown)


skovhus:
Is this an alternative to the recast parser suggested here https://github.com/facebook/jscodeshift/pull/239#issuecomment-367760997

domoritz:
Is anyone looking at this? It would be great to have support for Typescript.

fkling:
@domoritz Typescript support was just added in the latest version (v0.6.0). I picked the changes from #286 instead of this one (sorry @matthewrobertson).

PR #275, opened 2018-10-02
update babylon package to make it work

brcarey:


fkling:
Babylon has been updated to the relased v7 version. Not sure what "didn't work" for you, but hopefully  it does work now 😃

PR #277, opened 2018-10-10
hasAttributes filter method supports attributes with omitted boolean values

artemruts:
This solves an issue when `JSXElement`'s `hasAttribute` filter method can't find attributes where boolean value (`true`) was omitted.

E.g., given this `jsx`:

```jsx
<FooBar baz />
```
If I try to do something like this:

```js
const jsx = Collection.fromNodes(/* jsx nodes */)
  .findJSXElements()
  .filter(
    JSXElementCollection.filters.hasAttributes({
      baz: v => v === null
    })
  );
```
`JSXElement` will throw an error `TypeError: Cannot read property 'expression' of null`  since there is no `value`.

https://github.com/facebook/jscodeshift/blob/595b15cda4f52673cad8fe3ae65c12e334966dbf/src/collections/JSXElement.js#L100

balylon7 AST for above `jsx` code looks like this:
```js
{
  "type": "JSXAttribute",
  "start": 8,
  "end": 11,
  "loc": {
   /* ... */
  },
  "name": {
    "type": "JSXIdentifier",
     /* ... */
    },
    "name": "baz"
  },
  "value": null
}
```
___

I also ran `npm run docs` since I modified a few files, this produced lots of doc updates 😨 
Main changes are in `JSXElement.js` and `JSXElement-test.js` files.


artemruts:
@fkling Felix, I know you're busy but can I please get any feedback on this or is there someone who can review this PR?

Thanks.

PR #282, opened 2018-10-31
Add support for explicit flow type arguments

wanderley:
In order to support explicit flow type arguments, this PR upgrades the babylon parser from babylon 6 to @babel/parser ^7.0.0.

Example of new syntax:
```js
// @flow
const a = b<stirng>();
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "babel-preset-es2015": "^6.9.0",
     "babel-preset-stage-1": "^6.5.0",
     "babel-register": "^6.9.0",
-    "babylon": "^7.0.0-beta.47",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "babel-preset-es2015": "^6.9.0",
     "babel-preset-stage-1": "^6.5.0",
     "babel-register": "^6.9.0",
-    "babylon": "^7.0.0-beta.47",
+    "@babel/parser": "^7.0.0o",
```
TheSavior: This seems like a weird version number. Typo?

xixixao:
```suggestion
    "@babel/parser": "^7.0.0",
```

PR #284, opened 2018-11-05
SyntaxError-Using-the-export-keyword-between-a-decorator-and-a-class-…

amitksingh1490:
SyntaxError: Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead 
support for legacy decorators

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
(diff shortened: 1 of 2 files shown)


fkling:
Because it's not possible to support every configuration that someone needs `--parser-config` now allows you do pass your own configuration to babylon. Use that instead to enable decorators before exports behavior (https://babeljs.io/docs/en/babel-parser#options).

amitksingh1490:
That's great. Thanks

PR #287, opened 2018-11-19
Upgrade to @babel/parser; Apply Babel to TS transform files

brieb:
- Upgrade `babylon` ([deprecated](https://github.com/babel/babylon#babylon-has-been-moved-into-the-main-babel-mono-repo-as-babelparser)) to `@babel/parser`
- Upgrade other babel deps
- Support transpiling TS transform files via `@babel/preset-typescript`
  - For example, `jscodeshift -t sample/reverse-identifiers.ts all-the-code`

Reviewers: @fkling @cpojer


src/__tests__/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "globals": {
+    "jest": true
+  },
+  "env": {
+    "jasmine": true
+  }
+}
```
brieb: Converted this file from yml to json for consistency

src/Worker.js
```diff
@@ -53,15 +54,21 @@ function prepareJscodeshift(options) {
 
 function setup(tr, babel) {
   if (babel === 'babel') {
-    require('babel-register')({
+    require('@babel/register')({
       babelrc: false,
       presets: [
-        require('babel-preset-es2015'),
-        require('babel-preset-stage-1'),
```
brieb: `@babel/preset-stage-1` is deprecated. See https://babeljs.io/docs/en/babel-preset-stage-1
It was added in https://github.com/facebook/jscodeshift/commit/28eb51508dca2aa70f9a454d3247ee2e08fcf4ae.
Are there any experimental features we'd like to include?

fkling: Good question. I'll take a look at the current proposal.

Another question though: Can `@babel/preset-flow` and `@babel/preset-typescript` be both loaded at the same time? It feels like this should be incompatible (at least you cannot set both options in babylon).
cc @hzoo

hzoo: Yeah, you shouldn't be able to because they would both try to set options in babylon and fail as incompat

brieb: Oh interesting. Thank you so much for pointing this out! I didn't get an error from babel register about it and it appeared to run fine on the TS file I tested. But, I realize that my test TS file was valid flow so I didn't see any issues. I tried adding a TypeScript-only construct and it does indeed fail. I updated this PR to only load the typescript preset if it's a `.tsx?` transform.

package.json
```diff
@@ -24,11 +24,15 @@
...
-    "babel-register": "^6.9.0",
+    "@babel/core": "^7.1.6",
+    "@babel/parser": "^7.1.6",
+    "@babel/plugin-proposal-class-properties": "^7.1.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/preset-env": "^7.1.6",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-typescript": "^7.1.0",
+    "@babel/register": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
fkling: Is this needed or can I just remove it? @brieb

hzoo: it's needed if as a dep (it just links to `@babel/core`) for any dependency that uses `babel-core` as a dep like jest and hasn't explicitly upgraded to `@babel/core` themselves (https://github.com/facebook/jest/tree/master/packages/babel-jest#usage)

fkling: Ok, thank you @hzoo!

PR #291, opened 2018-12-04
Use @babel/parser in TypeScript parser

elliottsj:
Previously, the TypeScript parser would fail due to the 'babylon' package having the incorrect version (v6). Added a test which will detect this:

```
  ● jscodeshift CLI › --parser=ts › parses TypeScript sources

    expect(string).not.toContain(value)

    Expected string:
      "Processing 1 files...
     ERR /var/folders/ms/t3l222rn1v3dkvvnytj4vfkw0000gn/T/f-118113-6659-hurec.taoyh Transformation error (Unexpected token, expected ; (1:5))
    SyntaxError: Unexpected token, expected ; (1:5)
```

Switching to '@babel/parser' fixes this.

### DIFF
--- a/bin/__tests__/jscodeshift-test.js
+++ b/bin/__tests__/jscodeshift-test.js
@@ -321,7 +321,27 @@ describe('jscodeshift CLI', () => {
         }
       );
     });
-  })
+  });
+
+  describe('--parser=ts', () => {
+    it('parses TypeScript sources', () => {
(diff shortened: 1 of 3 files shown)


fkling:
Ugh, thanks for catching that! I remember doing these changes but ... ¯\_(ツ)_/¯ .

Will release a fixed version asap.

PR #296, opened 2018-12-11
VariableDeclarator: don't rename ClassProperty 

henryqdineen:
`renameTo` would mistakenly rename `ClassProperty` nodes and this fixes that.

I have also noticed that when using `bablyon` parser we would need support for ignoring `ObjectProperty`, `ObjectMethod` and `ClassMethod` types. Would that make sense to add?


fkling:
> Would that make sense to add?

We probably have to, yes.

PR #297, opened 2018-12-14
Introduce jest snapshot testing utils (fixes #135)

### LINKED ISSUE #135: Use Jest snapshot testing in testUtils
Jest 14 adds snapshot testing: https://facebook.github.io/jest/blog/2016/07/27/jest-14.html. One of the advantages is that it allows snapshots to be automatically updated by running `jest -u`.

Currently we manually load the expected result and compare with `toEqual`: https://github.com/facebook/jscodeshift/blob/master/src/testUtils.js#L68. I think this could be replaced with snapshot testing, avoiding the need for us to manually load the file containing the expected result.

dogoku:
The main benefit of this PR is actually the `applyTransform` function, since it can be used for a lot of things.

- Refactored `runInlineTest` and extracted `applyTransform` function
- Introduced `runSnapshotTest`, same as `runInlineTest` but takes no `expectedOutput`
- Introduced `defineSnapshotTest`, same as `defineInlineTest` but takes no `expectedOutput`
- `runInlineTest` and `runSnapshopTest` now both also return the result `applyTransform`
    - This can be used to take the result of one run and use it for another test (check for idempotent)

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,7 @@ jscodeshift comes with a simple utility to allow easy unit testing with [Jest](h
  - Test fixtures are located in a `__testfixtures__` directory
 
 This results in a directory structure like this:
+
 ```
 /MyTransform.js
 /__tests__/MyTransform-test.js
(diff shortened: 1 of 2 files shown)


ericwooley:
Any chance for a bump on this. It's been hanging for quite a while, and would make life so much easier.

Daniel15:
I can try to review this soon! In the meantime you might find the existing test utils useful.

Hypnosphi:
I've published this fork as a temporary scoped package: https://www.npmjs.com/package/@hypnosphi/jscodeshift

Hypnosphi:
@jbrown215 does `0.7.0` include this? If so, can you please mention it in the [changelog](https://github.com/facebook/jscodeshift/releases/tag/v0.7.0)?

Daniel15:
Looks like it's included: https://github.com/facebook/jscodeshift/compare/v0.6.4...v0.7.0

jbrown215:
@Hypnosphi: Yes and yes

Daniel15:
For what it's worth, I was just writing a transform today and really like that you separated out `applyTransform` method, because it means I can use it for inline snapshots (which I prefer over 'regular' snapshots). Thanks! :D

```js
const {applyTransform} = require('jscodeshift/dist/testUtils');
const fooToBar = require('../fooToBar');

function transform(input, options = null) {
  return applyTransform(fooToBar, options, {source: input});
}

test('first', () => {
  expect(
    transform(`foo()`),
  ).toMatchInlineSnapshot(`"bar();"`);
});
```

PR #300, opened 2019-01-10
bugfix: Avoid crash when there are no matching files

JHilker:
Currently in the case of no files, the first promise of the runner returns undefined, however this leads to a `Cannot read property 'Symbol(Symbol.iterator)' of undefined` when you pass that undefined value into `Promise.all(pendingWorkers)`.

The simple fix here is to return an empty array instead here so that `Promise.all([])` doesn't crash.

Node version: v8.9.1

Before:
image

After:
image


fkling:
#295 already provided that fix, I just didn't get around to merging it. Thanks though!

PR #308, opened 2019-02-27
Remove transform file .js restriction for testUtils

pengx17:
Since we are able to write transformers in TS, the .js extension hard coded in testUtils not allow us to use it to write tests for TS transform modules.
And `require('some-module')` will automatically infer some-module's extension whether it is ended with '.js' or '.ts', it should be safe to remove ',js'.


fkling:
Seems reasonable.

PR #310, opened 2019-03-08
support change default parser in fromSource

ifibercc:
file: core.js
method: fromSource

jscodeshift has many different parsers with useful plugins, but the fromSource method cannot use it .
It's a bother thing to define my own plugin collections for development. so support it.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -81,6 +81,9 @@ function fromSource(source, options) {
   if (!options.parser) {
     options.parser = getParser();
   }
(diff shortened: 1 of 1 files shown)


fkling:
You should be able to use `withParser` method like so:

    jsocdeshift = jscodeshift.withParser('flow');

though I agree that it should be possible to pass the parser as option to `jscodeshift`. But just updating `fromSource` won't be enough unfortunately.

Issue #316, opened 2019-04-09
Errors when running transform file from URL


felipemsantana:
When I try to use a remote transform file it throws a lot of errors related to ES modules:

```
➜ npx jscodeshift -t 'https://raw.githubusercontent.com/reactjs/react-codemod/master/transforms/manual-bind-to-arrow.js' src/**/*.jsx
npx: instalou 276 em 10.734s
Processing 36 files...
Spawning 3 workers...
Sending 12 files to free worker...
Sending 12 files to free worker...
Sending 12 files to free worker...
/tmp/jscodeshift11939-8547-1x2kkqo.acev:24
export default function transformer(file, api) {
^^^^^^

SyntaxError: Unexpected token export
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Module._extensions..js (module.js:664:10)
    at Object.newLoader [as .js] (/home/fsantana/.npm/_npx/8547/lib/node_modules/jscodeshift/node_modules/pirates/lib/index.js:104:7)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
/tmp/jscodeshift11939-8547-1x2kkqo.acev:24
```

But saving the file locally and running:
```
npx jscodeshift -t manual-bind-to-arrow.js src/**/*.jsx
```

Works correctly.

PR #317, opened 2019-04-20
fix: preserve file extension for remote file

samselikoff:
There is behavior in this program that depends on the extension
of the transformFile, notably how babel runs (or doesn't).

There was a bug when requesting a remote transformFile authored as
an ES6 module because of the temp filename being used.

This commit preserves the original file extension used when writing
the remote file to the temp one.

Fixes https://github.com/facebook/jscodeshift/issues/316

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -184,7 +184,8 @@ function run(transformFile, paths, options) {
             contents += d.toString();
           })
           .on('end', () => {
-            temp.open('jscodeshift', (err, info) => {
+            const ext = path.extname(transformFile);
+            temp.open({ prefix: 'jscodeshift', suffix: ext }, (err, info) => {
               if (err) return reject(err);
               fs.write(info.fd, contents, function (err) {
                 if (err) return reject(err);


samselikoff:
This is my first contribution to this project - let me know if I'm missing anything in my commit!

src/Runner.js
```diff
@@ -184,7 +184,9 @@ function run(transformFile, paths, options) {
             contents += d.toString();
           })
           .on('end', () => {
-            temp.open('jscodeshift', (err, info) => {
+            const re = /(?:\.([^.]+))?$/;
+            const ext = re.exec(transformFile)[1];
```
fkling: As far as I can tell, you should be able to use [`path.extname`](https://nodejs.org/api/path.html#path_path_extname_path) here. Then you don't have to deal with regular expressions.

samselikoff:
Nice, didn't know about that! Updated.

rajasegar:
@samselikoff Is the error you are talking about is the same as #316

samselikoff:
Yep, this should be the fix!

CvX:
It does indeed fix #316! 🙂
Anyone interested can use this PR via npx:

```bash
npx samselikoff/jscodeshift#fix-remote-transform-file -t https://url.here/codemod.js tests
```

PR #321, opened 2019-05-16
Bump recast to ^0.18.1

sodatea:
Diff here: https://github.com/benjamn/recast/compare/v0.16.1...v0.18.1


valerybugakov:
@mroch should this update be released?

PR #324, opened 2019-06-13
dynamic import for typescript transform files

mikob:


mikob:
For posterity, this screws up `  const module = require(tr);` by making it async, I don't need the functionality anymore so closing this.

Issue #254, opened 2018-04-12
Asynchronous code in transformers


akiran:
Is it possible to support asynchronous code in the transformers.

Our transformations depend on the data from database. We couldn't add asynchronous  code in transformers. I added async on the transformer function to use await inside transformer code. 
But it didn't work

sibelius:
have you figure it out how to solve this?

akiran:
@sibelius No, I didn't figure this out

mattkime:
I'd also like to see this. I'm using https://github.com/Galooshi/import-js to resolve import statements but this isn't possible due to a lack of async function support.

danieldelcore:
Hey all, I think this issue be closed now since merging this PR: https://github.com/facebook/jscodeshift/pull/237

PR #237, opened 2018-02-06
Allow transform to be a promise. Resolves #210.

rektide:
Hello. Wanted to float this, as a way to allow transforms to be async. This unlocks any kind of transform that would want to query some exterior piece of state.

PR #325, opened 2019-06-13
Async

mikob:
inspired by rektide/jscodeshift-async

doesn't have branch pollution

solves #254

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "write-file-atomic": "^2.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


danilovaz:
@gkz is this still something can be merged? I would love to help if need something to get this works

Daniel15:
I think this is a duplicate of #237.

Daniel15:
I don't think this is needed since #237 was merged a while back. Please let me know if that's not the case.

PR #329, opened 2019-06-20
Dependency updates

mfeckie:
Updates babel dependencies to allow newer syntax to be transformed.

This currently results in a babel parsing error, but is valid.

```
class Foo {
  @decoratedGenerator
  *doStuff() {}
}
```

With the upgraded dependencies this example can be transformed.

In babel 7.4.5 duplicate identifiers cannot be used, which mean that the test which declares a function named `func` twice is no longer valid.  

This can be seen in the babel REPL here https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABKSAKAlIg3gXwFB4rTxJEbY5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.4.5&externalPlugins=


wdoug:
I believe if this is updated further to bring the babel dependencies up to ^7.9.0 or higher this would resolve issue https://github.com/facebook/jscodeshift/issues/409

Daniel15:
Superseded by #410. Thanks for the contribution!

PR #332, opened 2019-07-14
Support different parsers in the test utils

skovy:
## Motivation

`jscodeshift` now supports TypeScript (and other parsers) via the CLI. The `react-codemod` package can now support different React codebases. 

## Changes

In order to use the `defineTest` helper in `testUtils.js` with different parsers the `defineTest` function needs to allow dynamic parsers and look at different extensions (for TypeScript: `.ts`, `.tsx`).

The approach used in this PR was to add another parameter `testOptions` that allows passing in an object with a `parser` key. The extension then can be inferred from the file being parsed.

I didn't feel great about adding another parameter so I opted for an object so if there are additional config options they can be added to this object rather than adding more arguments and passing many `null`s to get to the proper position. I also considered folding `testFilePrefix` into this object:

```typescript
{
  testFilePrefix?: string;
  parser?: "flow" | "ts" | "tsx" | ...;
}
``` 

However, this would be a breaking change for those using the `testUtils`. I'm happy to make that change but opted for avoiding the breaking change.

I also originally experiment with reusing the second existing `options` parameter but that's what get's passed to the transformer and it felt like mixing two unrelated option configs. Open to feedback on this approach. See the proposed pull request in `react-codemod` for how this helper would be used and the overall motivation for this change: https://github.com/reactjs/react-codemod/pull/228

## Testing

- See the new TypeScript sample added (in the `sample` directory)

### DIFF
--- a/sample/__testfixtures__/typescript/reverse-identifiers.input.ts
+++ b/sample/__testfixtures__/typescript/reverse-identifiers.input.ts
@@ -1,0 +1,5 @@
+const firstWord = 'Hello ';
+const secondWord = 'world';
+const message = firstWord + secondWord;
+
+const getMessage = (): string => message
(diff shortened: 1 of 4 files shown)


Daniel15:
Seems like a reasonable idea. Could you please rebase?

skovy:
Thanks for taking a look @Daniel15! This should be ready for another look 👌

vagusX:
Sorry but I have to ask that when will this feature release to npm?

PR #334, opened 2019-07-18
Fix Unhandled Promise rejection when called with a non-existent path

threepointone:
When jscodeshift is called with a path that doesn't exist, `concatAll()` gets called with `undefined`, which isn't an iterable, and throws an error. This PR fixes the promise being resolved by matching it with the ignore case.


AndersDJohnson:
Yes! I would love for this fix to be merged in and released, and/or the PR I raised (#339) to address the same symptom. Thanks!

PR #372, opened 2020-04-23
add pattern into addIgnorePattern

mickaelzhang:
I believe this is closer to how `.gitignore` work but I'll let you guys decide if it should be merged or closed since it change the current behaviour.

For example, if we have the following in `.gitignore`:
```
node_modules
coverage
```

It ignores both directory.

### DIFF
--- a/src/ignoreFiles.js
+++ b/src/ignoreFiles.js
@@ -15,6 +15,7 @@ function addIgnorePattern(val) {
   if (val && typeof val === 'string' && val[0] !== '#') {
     let pattern = val;
     if (pattern.indexOf('/') === -1) {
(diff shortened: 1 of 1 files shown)


gkz:
We aren't looking at making feature additions to jscodeshift

ElonVolo:
Just a note:  __—gitingore__ feature’s available in evcodeshift if anyone needs it.

ElonVolo:
@gkz I'm assuming that this is still the case.

ElonVolo:
@morgante And there we go.

gkz:
This is what we're using/maintaining these days: https://www.npmjs.com/package/hermes-transform

PR #375, opened 2020-05-03
Make @babel/preset-env optional, drop support for Node 8 and below

cpojer:
`@babel/preset-env` is a very large dependency to determine which plugins to use in node. In this PR I'm making it optional, dropping support for Node 8 and below (which have been deprecated for years) and added common syntax transforms to make jscodeshift transforms run in Node 10+.

The behavior will be unchanged if `@babel/preset-env` is already part of node_modules, otherwise it needs to be installed by the user. This will be released as jscodeshift 0.8 to make it non-breaking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - 6
-  - 8
   - 10
+  - 12
   - stable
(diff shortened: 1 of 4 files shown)


cpojer:
Actually learned that `optionalDependencies` is only for deps that can fail to install, but they will still be installed. I changed it to `peerDependencies` which prints a warning, I think that's a good middle ground.

https://github.com/facebook/jscodeshift/commit/2256c1d38c117c85e4f4872ea2ec66abe2765d10

PR #405, opened 2020-12-21
feat: create template.asyncExpression

jedwards1211:
allows things like `` asyncExpression`foo ? await bar : baz` ``

I was tempted to make `expression` itself support this, but it has to wrap the code in `async () => (...)`, and theoretically some user-supplied parsers might not support async functions.


Daniel15:
Published this in 0.13.0. Thanks for your contribution!

PR #407, opened 2021-01-21
chore: update recast to allow ChainExpression types

lpoulter:
Currently jscodeshift errors when it encounters [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

`Error: did not recognize object of type "ChainExpression"`

Ast-types and reacast where [recently updated](https://github.com/benjamn/ast-types/pull/399) to support ChainExpressions this pr bump fixes jsShift errors on chainexpressions

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "micromatch": "^3.1.10",
     "neo-async": "^2.5.0",
     "node-dir": "^0.1.17",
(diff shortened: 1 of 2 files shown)


gkz:
Thanks. You need to update the lockfile as well

PR #410, opened 2021-01-26
Support newer TypeScript syntax

### LINKED ISSUE #409: Syntax error on TypeScript's import type syntax
Currently, jscodeshift fails to parse TypeScript 3.8's [`import type` syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) when using the `tsx` parser. A simple demo of this can be seen in [this AST Explorer gist](https://astexplorer.net/#/gist/73bbae5a9084d88eb0f4fa0b4e97f8fe/06b039e8636be4f455afb639fa16d1e9fcd8964f) here.

I believe the issue is caused by the fact that the root version of `@babel/parser` used in this project is [7.2.0](https://github.com/facebook/jscodeshift/blob/48f5d6d6e5e769639b958f1a955c83c68157a5fa/yarn.lock#L236) (this is the [parser used for `tsx`](https://github.com/facebook/jscodeshift/blob/48f5d6d6e5e769639b958f1a955c83c68157a5fa/parser/tsx.js#L12-L27)). The version needed to support TypeScript 3.8 syntax, however, is version [7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0).

wdoug:
This builds on top of @mfeckie's pull request https://github.com/facebook/jscodeshift/pull/329 to update babel dependencies to support newer syntax updates. 

This closes #409


lencioni (approved):

wdoug:
I think there was an issue with the Travis-CI run. It says it failed, but the logs are all null...

TheMcMurder:
Is the travis failure related to travis-ci.org shutdown?

wdoug:
@Daniel15, any possibility we could work to get this merged? This is currently blocking us (and presumably many others) from being able to use jscodeshift

Daniel15:
Sure! This looks reasonable to me. However, it looks like there's some merge conflicts. @wdoug If you rebase this and ping me again, I can merge it :)

>  This is currently blocking us (and presumably many others) from being able to use jscodeshift

Internally at Facebook we use the Flow parser which is probably why nobody has noticed this.

wdoug:
Hmm. @Daniel15, I believe I already resolved those... at least GitHub is telling me there are no conflicts: 
Screen Shot 2021-04-21 at 8 06 23 AM

Maybe your page was cached from earlier?

Daniel15:
@wdoug Huh, that's weird, because I still see this today (on a different PC so it shouldn't be caching):

image

wdoug:
Hmm. @Daniel15 do y'all only support rebase and merge (as opposed to merge commits or squash and merge)? If so, I could see how that would be a problem given that I merged in changes on `master` instead of rebasing them. I was planning on keeping it that way to keep the original commits that @mfeckie wrote, but if need be I can squash the commits on this branch if we think that might resolve the issue.

Daniel15:
Ah, good point, I think I can use a merge commit for this. Let me see...

wdoug:
I can squash my commits if you would like

wdoug:
@Daniel15 how do y'all manage your release process for this project? Just wondering when we might be able to update the version dependency for jscodeshift in things like `ts-migrate` ([related issue](https://github.com/airbnb/ts-migrate/issues/91))?

Daniel15:
I'll push an update now 😃

Daniel15:
Published as 0.12.0. Seems like the publish worked properly to me, but please let me know if anything looks amiss. Thanks for your contribution!

Daniel15:
@wdoug That's likely because of caching 😃  Yarn just pulls from npm.

wdoug:
Okay, cool. I thought they came from the same place but wasn't sure if something changed there. Thanks!

PR #416, opened 2021-03-05
Add a flag to make jscodeshift return a 1 code when errors were found

### LINKED ISSUE #194: [Bug] Exit code always 0
jscodeshift executions always exits with code `0`, regardless of whether there were any errors. Here is my output, although this is generic to any situation where you have a non-zero `errors` count.

```
Processing 1 files...
Spawning 1 workers...
Sending 1 files to free worker...
 ERR /var/folders/tv/fdtq5cgn6zq4bj1_lzvrhl7h0000gn/T/07741760665/__testfixtures__/replace-require.input.js Transformation error
Error: arg "toReplace" required. Received undefined
    at req (/Users/dawsonbotsford/code/uber-codemods/src/replace-require.js:69:11)
    at replace (/Users/dawsonbotsford/code/uber-codemods/src/replace-require.js:74:3)
All done.
Results:
1 errors
0 unmodified
0 skipped
0 ok
Time elapsed: 0.688seconds
```

```sh
 ~/code/uber-codemods   regex-binding-bug ●  echo $?
0
```

Would the best practice instead be to set an exit code that's non-zero if there are any `errors`? In the current implementation, it's not intuitive to create an `exec` wrapper or the like in order to execute jscodeshift programmatically.

marcodejongh:
Fixes #194

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -298,6 +298,10 @@ function run(transformFile, paths, options) {
             process.stdout.write(
               'Time elapsed: ' + timeElapsed + 'seconds \n'
             );
+            
+            if (options.failOnError && fileStats.error > 0) {
+              process.exit(1);
+            }
           }
           if (usedRemoteScript) {
             temp.cleanupSync();
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -299,7 +299,7 @@ function run(transformFile, paths, options) {
               'Time elapsed: ' + timeElapsed + 'seconds \n'
             );
             
-            if (options.failOnError && fileStats.error > 0) {
+            if (options.failOnError && fileCounters.error > 0) {
               process.exit(1);
             }
           }
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ All options are also passed to the transformer, which means you can supply custo
   -v, --verbose=0|1|2           show more information about the transform process
                                 (default: 0)
       --version                 print version and exit
+      --fail-on-error           return a 1 exit code when errors were found during execution of codemods
 ```
 
 This passes the source of all passed through the transform module specified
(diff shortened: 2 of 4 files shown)


RajaBellebon (approved):

marcodejongh:
@gkz or @fkling Any chance we could get this merged?

src/Runner.js
```diff
@@ -298,6 +298,10 @@ function run(transformFile, paths, options) {
             process.stdout.write(
               'Time elapsed: ' + timeElapsed + 'seconds \n'
             );
+            
+            if (options.failOnError && fileCounters.error > 0) {
+              process.exit(1);
```
Daniel15, line 303: This should probably go below the `temp.cleanupSync` call at least?

Is it possible to put the `process.exit(1)` call in `bin/jscodeshift.js` rather than in this file? I suspect this file is supposed to be reusable as a library and thus shouldn't be handling the exit here.

marcodejongh: Good point on moving the process.exit(1), it also didn't really belong in the `if (!options.silent) {` statement.

On moving this to the bin, I think it makes sense to have it in this file since Runner.js already does many process stuff (see also the process.stdout). If you want to use this file in other JS code, the user can not pass in the failOnError flag.

But it looks like it passes back the stats to bin/jscodeshift, so it's easy to move there as well, it was just that bin/jscodeshift seemed like just the cli wrapper

marcodejongh:
@Daniel15 Any chance this could get another review? I'd love to be able to move us off of our fork of this library

Daniel15:
@marcodejongh  Sorry for the delay! I'll take a look at this tomorrow morning :)

Daniel15:
Looks good. Thanks!

RajaBellebon:
Hello @Daniel15 , thanks for accepting this pull-request! Do we have an approx idea of when the new version will be released?

Thank you!

Daniel15:
I'll try to push a new release tomorrow :)

Daniel15:
Sorry for the delay. Published this as 0.13.0. Thanks for your contribution!

PR #427, opened 2021-04-29
Remove lodash dependency from tsx parser

### LINKED ISSUE #418: Tsx parser imports hoisted lodash version
[This line](https://github.com/facebook/jscodeshift/blob/5910a4dcde431813c2e34a3a6f47e8800d275eb6/parser/tsx.js#L11) can cause transforms that have a second transitive dep on lodash to crash unexpectedly when lodash is unhoisted. I think this could be addressed a couple ways- either jscodeshift could express a dep on lodash or the tsx parser could remove the reference to lodash. I lean towards removing the lodash usage entirely (or at least not importing all of lodash to call one util one time).

### LINKED ISSUE #426: TSX parser does not work if lodash is not already installed
## Sample project

```tsx
// thing.tsx
import React from "react";

export const Mine = () => (
    <h1>Hi</h1>
);
```

```js
// transform.js
module.exports = function(fileInfo) {
	return fileInfo.source;
};
```

## STR

1. `npm install react`
2. `npx jscodeshift --parser=tsx thing.tsx`

## Result

```
npx: installed 230 in 8.925s
Processing 1 files... 
Spawning 1 workers...
Sending 1 files to free worker...
 ERR thing.tsx Transformation error (Cannot find module 'lodash' Require stack: - /Users/me/.npm/_npx/15730/lib/node_modules/jscodeshift/parser/tsx.js - /Users/me/.npm/_npx/15730/lib/node_modules/jscodeshift/src/getParser.js - /Users/me/.npm/_npx/15730/lib/node_modules/jscodeshift/src/Worker.js)
Error: Cannot find module 'lodash'
Require stack:
- /Users/me/.npm/_npx/15730/lib/node_modules/jscodeshift/parser/tsx.js
- /Users/me/.npm/_npx/15730/lib/node_modules/jscodeshift/src/getParser.js
All done. 
Results: 
1 errors
0 unmodified
0 skipped
0 ok
Time elapsed: 0.496seconds
```

From what I can tell, all that needs to happen is to add `lodash` to this project's `dependencies`. (PR incoming)

robyoder:
Lodash is required by the `tsx` parser, but it was not listed in `dependencies`, so `npx jscodeshift --parser=tsx thing.tsx` would always fail. 

Resolves #426
Resolves #418 

I added the dependency as `^4.17.19` because that was one that was already included in yarn.lock (thanks to existing dev dependencies), so as to avoid making any bigger diff than necessary.

**Update:** see comments below. I removed the lodash dependency entirely instead.

### DIFF
--- a/parser/tsx.js
+++ b/parser/tsx.js
@@ -8,11 +8,9 @@
 
 'use strict';
 
-const _ = require('lodash');
 const babylon = require('@babel/parser');
-const baseOptions = require('./tsOptions');
-
-const options = _.merge(baseOptions, { plugins: ['jsx'] });
+const options = require('./tsOptions');
+options.plugins.push('jsx');
 
 /**
  * Doesn't accept custom options because babylon should be used directly in
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "colors": "^1.1.2",
     "flow-parser": "0.*",
     "graceful-fs": "^4.2.4",
+    "lodash": "^4.17.19",
     "micromatch": "^3.1.10",
     "neo-async": "^2.5.0",
     "node-dir": "^0.1.17",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "colors": "^1.1.2",
     "flow-parser": "0.*",
     "graceful-fs": "^4.2.4",
-    "lodash": "^4.17.19",
     "micromatch": "^3.1.10",
     "neo-async": "^2.5.0",
     "node-dir": "^0.1.17",


robyoder:
I’m certainly not opposed. I was surprised to find the lodash dependency for a single line like that, but I figured this would be the least controversial change. I’ll touch it up.

robyoder:
Great call. There was actually some pretty unexpected behavior that resulted from the way it was being used. `_.merge` doesn't concatenate arrays like one might have expected from looking at the code:

merge

So in this case, the first plugin was actually getting replaced with `'jsx'`, probably not what was intended.

`_.merge` does mutate the first parameter, so I went ahead and just pushed the extra plugin onto the array.

Thanks for the poke to reconsider. :)

parser/tsx.js
```diff
@@ -8,11 +8,9 @@
 
 'use strict';
 
-const _ = require('lodash');
 const babylon = require('@babel/parser');
-const baseOptions = require('./tsOptions');
-
-const options = _.merge(baseOptions, { plugins: ['jsx'] });
+const options = require('./tsOptions');
+options.plugins.push('jsx');
```
Daniel15, line 13: This will mutate `baseOptions.plugins`, since it's the same array reference, which I don't think is what we want.

I think you want to instead clone `baseOptions` and add the `jsx` plugin... Something like this:
```js
const baseOptions = require('./tsOptions');
const options = {
  ...baseOptions,
  plugins: [...baseOptions.plugins, 'jsx'],
);
```

However we might have to use basic JS here (ie. no object spreads) for compatibility with older Node.js versions, so maybe something like this:
```js
const baseOptions = require('./tsOptions');
const options = Object.assign({}, baseOptions);
options.plugins = ['jsx'].concat(baseOptions.plugins);
```

Please test it though :)

Daniel15 (changes requested):

parser/tsx.js
robyoder: As I mentioned above, the existing code was using `_.merge` which also mutated `baseOptions`.

Daniel15:
Merged in #432. Thank you!

PR #428, opened 2021-05-05
Add missing lodash dependency

marcodejongh:
Lodash is required by parser/tsx.js but is not explicitly depended on.
Usually it seems to work okay due to the fact that other dependencies of jscodeshift
pull in lodash, but for some reason it started failing recently for us when running
jscodeshift through npx, so figured it's time to correct the missing dep

See: https://github.com/facebook/jscodeshift/blob/master/parser/tsx.js#L11

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "colors": "^1.1.2",
     "flow-parser": "0.*",
     "graceful-fs": "^4.2.4",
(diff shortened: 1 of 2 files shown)


marcodejongh:
Nvm just found: https://github.com/facebook/jscodeshift/pull/427/files

PR #431, opened 2021-05-11
Add lodash as explicit dependency

raon0211:
JSCodeShift [uses lodash](https://github.com/facebook/jscodeshift/blob/57a9d9c731429da624655472290a7b4802e7e8d7/parser/tsx.js#L11) when using the tsx parser. 

However, lodash is not specified explicitly in package.json. It seems that it is available via implicit [phantom dependency](https://rushjs.io/pages/advanced/phantom_deps/).

This makes JSCodeShift to break when using strict package managers such as Yarn 2+ or pnpm.

This PR includes lodash as an explicit dependency to fix the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "colors": "^1.1.2",
     "flow-parser": "0.*",
     "graceful-fs": "^4.2.4",
+    "lodash": "^4.17.21",
     "micromatch": "^3.1.10",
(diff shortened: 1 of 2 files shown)


raon0211:
Will close the PR as an duplicate of https://github.com/facebook/jscodeshift/pull/427. 😓

PR #432, opened 2021-05-17
Remove lodash dependency from tsx parser 

### LINKED ISSUE #418: Tsx parser imports hoisted lodash version

JHilker:
Resolves #418
Replaces https://github.com/facebook/jscodeshift/pull/427

Opening a new PR with some added tests and the suggested fixes since the last PR has set for a few weeks now. Feel free to close this in favor of the PR added by @robyoder, but this bug is painful enough for my use case that I'd love to see it get fixed and released, so figured I'd quickly do some of the legwork for testing it out 😄.

The PR:
* Removes the undeclared dependency on lodash
* Stops overriding the value of `asyncGenerators` in the plugins array with `jsx`
* Stops mutating the return value of `tsOptions`

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "roots": [
       "src",
       "bin",
+      "parser",
       "sample"
     ]
   }
--- a/parser/__tests__/.eslintrc
+++ b/parser/__tests__/.eslintrc
@@ -1,0 +1,8 @@
+{
+  "globals": {
+    "jest": true
+  },
+  "env": {
+    "jasmine": true
+  }
+}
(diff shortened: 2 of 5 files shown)


Daniel15:
This looks good to me. Thanks @JHilker and @robyoder :)

Daniel15:
Published this in 0.13.0. Thanks for your contribution!

PR #433, opened 2021-05-19
Make dependency on temp version slightly stricter

JHilker:
With `temp@0.8.3` installed and the `babel` option enabled, Babel throws due to the legacy octal literals used within `temp/lib/temp.js`:

```
SyntaxError: /.../node_modules/temp/lib/temp.js: Legacy octal literals are not allowed in strict mode. (226:20)

  224 | function mkdir(affixes, callback) {
  225 |   var dirPath = generateName(affixes, 'd-');
> 226 |   fs.mkdir(dirPath, 0700, function(err) {
      |                     ^
  227 |     if (!err) {
  228 |       deleteDirOnExit(dirPath);
  229 |     }
```

It looks like this issue was fixed in `0.8.4`, which uses `parseInt` instead. Most packages would automatically be using this newer version regardless, but I ran across a few at my org where after upgrading `jscodeshift` I'd hit this issue since `temp` was locked to an older version. Making the dep a bit more explicit here should save needing to manually upgrade the transitive dependency in these cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "neo-async": "^2.5.0",
     "node-dir": "^0.1.17",
     "recast": "^0.20.4",
-    "temp": "^0.8.1",
+    "temp": "^0.8.4",
     "write-file-atomic": "^2.3.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


Daniel15:
Thanks!

Issue #294, opened 2018-12-05
Improve/fix applying babel to the transformer


fkling:
**Note:** If you have problems running codemods and get Babel related errors, run jscodeshift with `--no-babel`.

---

The last change to `src/Worker.js` causes issues with `babel-register`. Passing module identifiers causes `babel-register` to look up those modules in the wrong places. Those modules should be `require`d instead.

Furthermore, we probably don't want to transpile most of ES2015 features since people probably use a Node version that supports those. `present-env` already supports this:

> If you want to compile against the current node version, you can specify "node": true or "node": "current", which would be the same as "node": process.versions.node.

https://babeljs.io/docs/en/babel-preset-env#targetsnode

but we should also include/exclude plugins based on the current node version.

brieb:
Thank you for fixing this. So sorry for introducing this issue. Will keep this in mind for future reference. TIL.

Should we also add `/\/node_modules\//` to this ignore list?
https://github.com/facebook/jscodeshift/commit/cb92c488b1fd068dbbd8f87ae380d2d8b3383513#diff-e957a0672425b8dcfdcfe9f8cccf36f7

PR #435, opened 2021-05-27
Worker: Babel ignore node_modules by default + some minor babel transform perf improvements

JHilker:
I ended up digging a bit into the performance here since I noticed that with newer versions of `jscodeshift`, what used to be quick (under Jests 5s timeout) was now consistently taking over 10 seconds. 

(Aside, it actually was taking closer to 30s when I first started, but I believe this was due to an insanely large default `BABEL_CACHE_PATH` that consistently took over 5s to read from disk. Deleting that was my biggest perf win by far)

For the purposes of my testing, I was running a jest test that also happened to run jscodeshift via the node API. This task also imported several larger dependencies. 

After the cache was cleared, via performance timing in the `setup` function, this function took an average of 14.157 seconds across 5 runs (13.661, 13.479, 13.438, 15.136, 15.071) with a hot cache and 11.408 seconds across 5 runs (12.133, 11.419, 11.321, 11.201, 10.965) with the cache disabled with the `BABEL_DISABLE_CACHE` env variable.

I'll probably continue to use `BABEL_DISABLE_CACHE` in my tooling going forward due to it seemingly being a performance improvement when `babel/register` is used like this, so all numbers going forward will make the assumption that it is being used.

The swap to require `plugins` before `babel/register` installed it hooks and reverting those hooks after the transform were required aren't too significant, but seemed worth including since I already had tested them locally. They reduced the average time to 10.571 seconds across 5 runs (10.769, 10.884, 10.379, 10.510, 10.312), a 7.3% decrease. Both these changes relate to just avoiding having `require` logic go through `babel/register` if it doesn't need to, either be requiring it before it hooks in, or removing the hooks before further `require`s occur.

---

The bigger, and more controversial change, is adding `/node_modules/` to the `ignore` list. This was proposed in https://github.com/facebook/jscodeshift/issues/294#issuecomment-452970300, but I'm not sure if there is more context there. [By default `babel/register` would ignore the `node_modules` of the cwd](https://github.com/babel/babel/blob/2a4f16236656178e84b05b8915aab9261c55782c/packages/babel-register/src/node.js#L146), but we explicitly want to compile files outside of the cwd here 🤔 . Should this be scoped to the transforms directory? Any `node_modules` in parent directories of the transform? All `node_modules`? Should it be configurable instead?

For my transform specifically, this reduced the `setup` time down to an average of 2.536 seconds across 5 runs (2.866, 3.169, 2.553, 1.632, 2.458), and additional 76% decrease from before. Though this is obviously dependent what dependencies a transform is pulling in

I'm happy to split the `/node_modules/` change into a separate PR for further discussion if desired, but figured I'd put it all here for now since they are a bit related.

### DIFF
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -52,6 +52,8 @@ function prepareJscodeshift(options) {
 }
 
 function setup(tr, babel) {
+  let babelRegister;
+
   if (babel === 'babel') {
     const presets = [];
     if (presetEnv) {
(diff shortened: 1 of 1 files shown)


src/Worker.js
```diff
@@ -89,6 +96,11 @@ function setup(tr, babel) {
   }
 
   const module = require(tr);
+
+  if (babelRegister) {
+    babelRegister.revert();
+  }
```
JHilker, line 102: This test: 

https://github.com/facebook/jscodeshift/blob/57a9d9c731429da624655472290a7b4802e7e8d7/bin/__tests__/jscodeshift-test.js#L194-L213

Catches that while reverting the babel register here works for any require/import that would be run at `require` time, it wouldn't catch any that would only be executed at run time.

This part isn't too significant, so supporting that case probably more important than worrying about reverting the register? Happy to 🔪

PR #471, opened 2022-01-04
feat: add a snapshot test util that loads the file for users

shriuken:
Adds a wrapper function, `defineSnapshotWithFixture` around `defineSnapshotTest` that mimics the loading/parsing behavior of `defineTest` to load the file based on the testFilePrefix in the `__testfixtures__` directory. Adds a quick test in the sample directory as well as updates documentation.


Daniel15:
Interesting! These days I usually use `toMatchInlineSnapshot` with my jscodeshift codemods, but I can see this being useful for larger test cases.

PR #485, opened 2022-02-16
Tabs to spaces in code examples for consistency

baweaver:
One of the code blocks was using tabs instead of spaces, which was inconsistent with the rest of the README.


Daniel15:
Thanks!

PR #497, opened 2022-03-26
chore(deps-dev): bump jest to 26.x

trivikr:
I attempted bumping to jest 27.x in https://github.com/trivikr/jscodeshift/pull/3 but there are [too many breaking changes](https://jestjs.io/blog/2021/05/25/jest-27#miscellaneous-breaking-changes).

The jest 27.x bump can be attempted after merging changes from evcodeshift fork.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.1",
     "eslint": "^5.9.0",
-    "jest": "^25",
+    "jest": "^26",
     "jsdoc": "^3.4.0",
     "mkdirp": "^0.5.1"
   },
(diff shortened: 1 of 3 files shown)


ElonVolo:
I actually coded a Jest 27 bump while working on jscodeshift itself (not the evcodeshift fork) about two months ago. In addition to the breaking changes in Jest itself, there was an issue that IIRC was caused by the jscodeshift Workers use of the module.parent property (which has since been deprecated in node.js) not playing nicely with Jest 27.

It turns out that the module.parent property is being used as a way to detect whether a transform is being run from a forked off process vs being run in-band (e.g if you're debugging, running unit tests, calling via jscodeshift API, etc). I ended up having to do a full refactoring of Worker.js to stop making use of module.parent and to instead use a more explicit method of determining whether transforms were being run. Since you're looking at bumping jest up to 27 I've started working on this branch again, but there's one or two kinks that still have to be ironed out.

PR #507, opened 2022-05-10
Removing dependency on args being in alphabetical order.

ElonVolo:
Previously, the arg parser would display the arguments in the help section in alphabetical order. If you run ```jscodeshift -h``` you've notice all the arguments are displayed in alphabetical order. The algorithm that prints out the help page actually relies on them being in alphabetical order. The problem with this approach is that if there's a command that should be included in a category of arguments that appears in the middle of the page, but the new argument starts with a "z", then the new argument will appear in the last lines on the help page, and not in the middle where it's supposed to be.

Since we're need to move in the direction of grouping arguments by functionality independent of names (like for upcoming --gitignore flag), we now have to come up with a quick refactoring that allows arguments to be displayed in a specified order regardless of what alphabetical order the arguments themselves occupy.

### DIFF
--- a/src/argsParser.js
+++ b/src/argsParser.js
@@ -48,7 +48,17 @@ function formatOption(option) {
 function getHelpText(options) {
   const opts = Object.keys(options)
     .map(k => options[k])
-    .sort((a,b) => a.full.localeCompare(b.full));
+    .sort((a,b) => {
+      if (a.display_index > b.display_index) {
+        return 1;
+      } else if (a.display_index < b.display_index) {
+        return -1;
+      } else if (a.display_index === b.display_index) {
+        return 0;
+      } else {
+        return -1;
+      }
+    });
 
   const text = `
 Usage: jscodeshift [OPTION]... PATH...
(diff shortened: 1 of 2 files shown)


ElonVolo:
@Daniel15

src/argsParser.js
```diff
@@ -48,7 +48,17 @@ function formatOption(option) {
...
+      if (a.display_index > b.display_index) {
+        return 1;
+      } else if (a.display_index < b.display_index) {
+        return -1;
+      } else if (a.display_index === b.display_index) {
+        return 0;
+      } else {
+        return -1;
+      }
+    });
```
Daniel15, line 61: Could you just do this:
```
.sort((a,b) => a.display_index - b.display_index)
```

Daniel15:
> like for upcoming --gitignore flag

Can we call that something more generic so that it could handle other ignore files (like `.hgignore`) too?

src/argsParser.js
ElonVolo: Changed

PR #508, opened 2022-05-17
Adding --gitignore flag

ElonVolo:
Adds a --gitignore flag that uses a .gitignore file in the current working directory to force jscodeshift to abstain from transforming files that match patterns specified in the .gitignore file.

Please note that this feature is added primarily for the benefit of beginners to jscodeshift, as a way to make the enormously timesaving technique of passing a project's .gitignore file to jscodeshift easily discoverable and learnable. The idea is that when the user looks at the jscodeshift help page, "--gitignore" jumps right out at them and they immediately make the connection.

It took me a year and a half of using jscodeshift before I realized I could simply just pass .gitignore to the utiliity, and creating a new ignore file for a freshly cloned project every day is an annoyance I wouldn't wish on others.

Due to the --gitignore flag being a discoverability feature, renaming it to something generic would impede the discoverability it's trying to promote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ All options are also passed to the transformer, which means you can supply custo
   -h, --help                    print this help and exit
       --ignore-config=FILE ...  ignore files if they match patterns sourced from a configuration file (e.g. a .gitignore)
       --ignore-pattern=GLOB ...  ignore files that match a provided glob expression
+      --(no-)gitignore          adds entries the current directory's .gitignore file
+                                (default: false)
       --parser=babel|babylon|flow|ts|tsx  the parser to use for parsing the source files
                                           (default: babel)
       --parser-config=FILE      path to a JSON file containing a custom parser configuration for flow or babylon
@@ -215,14 +217,14 @@ __Example: specifying parser type string in the transform file__
 module.exports = function transformer(file, api, options) {
   const j = api.jscodeshift;
   const rootSource = j(file.source);
-  
+
   // whatever other code...
-  
+
   return rootSource.toSource();
 }
-  
+
 // use the flow parser
-module.exports.parser = 'flow'; 
+module.exports.parser = 'flow';
 ```
 
 __Example: specifying a custom parser object in the transform file__
@@ -232,9 +234,9 @@ __Example: specifying a custom parser object in the transform file__
 module.exports = function transformer(file, api, options) {
   const j = api.jscodeshift;
   const rootSource = j(file.source);
-  
+
   // whatever other code...
-  
+
   return rootSource.toSource();
 }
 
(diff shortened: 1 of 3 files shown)


bin/__tests__/jscodeshift-test.js
```diff
@@ -20,6 +20,7 @@ const path = require('path');
 const temp = require('temp');
 const mkdirp = require('mkdirp');
 const testUtils = require('../../utils/testUtils');
+const { chdir } = require('process');
```
Daniel15, line 23: Nit: Remove spaces - ~~Facebook~~ Meta style has no spaces around destructured variables

```
const {chdir} = require('process');
```

I guess we should configure Prettier for jscodeshift one day.

README.md
```diff
@@ -394,6 +396,40 @@ jscodeshift(ast).findIdentifiers().logNames();
...
+
+__--ignore-config__ takes one or more paths to files containing lines with [.gitignore format](https://git-scm.com/docs/gitignore#_pattern_format) glob patterns.
+```
+
+// note: .gitignore is a random made-up filename extension for this example
+
+jscodeshift -—ignore-config="MyIgnoreFile.gitignore" mytransform.js
+
+// More than one ignore file
+jscodeshift -—ignore-pattern="first_ignore_file.gitignore” -—ignore-pattern="second_ignore_file.gitignore” mytransform.js
```
Daniel15, line 430: It looks like all the new example commands use a hyphen and a dash (`-—`) rather than two hyphens (`--`). Can you please fix those?

README.md
```diff
@@ -394,6 +396,40 @@ jscodeshift(ast).findIdentifiers().logNames();
...
+__--ignore-pattern__  takes a [.gitignore format](https://git-scm.com/docs/gitignore#_pattern_format) glob pattern that specifies file and directory patterns to ignore
+
+```
+jscodeshift -—ignore-pattern="js_configuration_files/**/*” mytransform.js
+
+// More than one ignore
+jscodeshift -—ignore-pattern="first_ignored_dir/**/*” -—ignore-pattern="second_ignored_dir/**/*” mytransform.js
+```
+
+__--ignore-config__ takes one or more paths to files containing lines with [.gitignore format](https://git-scm.com/docs/gitignore#_pattern_format) glob patterns.
```
Daniel15, line 422: So is `--ignore-config` just like `--gitignore` except you can specify the paths to the files with the ignore patterns?

Daniel15 (approved):
Looks pretty good to me. Thanks!

README.md
ElonVolo: Exactly. ```--ignore-config``` allows for more custom functionality than ```--gitignore__```. 

The problem was never that jscodeshift's ignore features didn't provide extensive custom functionality for advanced users; the problem was that jscodeshift's ignore functionality didn't provide any basic, easy-to-discover features for beginning users.

bin/__tests__/jscodeshift-test.js
ElonVolo: @Daniel15 fixed.

README.md
ElonVolo: Fixed.

Issue #512, opened 2022-07-15
feat: Is it possible to support full glob patterns without relying on bash


rxliuli:
Now nodejs itself already has a glob package, why use shell, which is not very friendly to windows users. Even when using GitBash, the `**` syntax is not supported by default. If I create a PR, will it be accepted?

Daniel15:
Sure, feel free to submit a PR. Please include unit tests if possible 😄

rxliuli:
I have already submitted a PR, can you help me review it?

PR #514, opened 2022-07-20
feat: Use fast-glob actually authentic glob road diameter match

rxliuli:
ref: https://github.com/facebook/jscodeshift/issues/512

Sorry, I didn't find the Runner.test.js unit test, so the unit test was not added.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -137,9 +137,7 @@ function getAllFiles(paths, filter) {
   return Promise.all(
     paths.map((file) => {
       if (file.includes('*')) {
-        return fg([file]).then((list) =>
-          list.filter((name) => filter(name) && !ignores.shouldIgnore(name))
-        );
+        return fg([file], { ignore: ignores.matchers, dot: true });
       }
       return new Promise((resolve) => {
         fs.lstat(file, (err, stat) => {
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -135,38 +135,33 @@ function dirFiles (dir, callback, acc) {
 
 function getAllFiles(paths, filter) {
   return Promise.all(
-    paths.map(
-      (file) =>
-        new Promise((resolve) => {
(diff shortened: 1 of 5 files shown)


rxliuli:
I have signed the cla, can I continue?

src/Runner.js
```diff
@@ -135,27 +135,38 @@ function dirFiles (dir, callback, acc) {
...
+      (file) =>
+        new Promise((resolve) => {
+          if (file.includes('*')) {
+            return fg([file])
+              .then((list) =>
+                list.filter(
+                  (name) => filter(name) && !ignores.shouldIgnore(name)
+                )
+              )
+              .then(resolve);
```
Daniel15, line 148: Instead of creating a new Promise and calling `resolve` in `then`, can you return the fast-glob promise directly? For example:
```
    paths.map(
      (file) =>
        if (file.includes('*')) {
          return fg([file])
            .then((list) =>
              list.filter(
                (name) => filter(name) && !ignores.shouldIgnore(name)
              )
            );
        }

        return new Promise((resolve) => {
          // ..... the old code here
  }
);
```

Daniel15 (commented):
Is it possible to write a unit test for this?

src/Runner.js
rxliuli: This has been processed. 

Need to test fast-glob? I have now modified to ignore files by fast-glob, ref

https://github.com/facebook/jscodeshift/pull/514/commits/e04ff64533ac86cf30540f1145553a0418c79bf5#diff-63495596a6a2d9c7803d0a490618a51baa6ab1178c89cdcc3f612f451e76c190R140

rxliuli:
> Is it possible to write a unit test for this?

I added some unit tests for getAllFiles

rxliuli:
Is there any problem preventing this PR from being merged now? @Daniel15

PR #515, opened 2022-07-25
Allow the j shortcut in testUtils

no23reason:
Since the API allows the `j` shortcut, the mock test API should implement it as well.


Daniel15:
Thanks!

PR #518, opened 2022-07-27
fix: async transform test support

### LINKED ISSUE #516: defineTest can not resolve async transform
If the transform is async, defineTest run error, because the defineTest only support sync, output is promise
image

shiyangzhaoa:
fix #516

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.1",
     "eslint": "^5.9.0",
+    "is-promise": "^4.0.0",
     "jest": "^26",
     "jsdoc": "3.6.7",
     "mkdirp": "^0.5.1"
--- a/sample/reverse-identifiers-async.js
+++ b/sample/reverse-identifiers-async.js
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Example jscodeshift async transformer. Simply reverses the names of all
+ * identifiers.
+ */
+ function transformer(file, api) {
+  const j = api.jscodeshift;
+
+  return Promise.resolve(
+    j(file.source)
+      .find(j.Identifier)
+      .replaceWith(
+        p => j.identifier(p.node.name.split('').reverse().join(''))
+      )
+      .toSource()
+  );
+}
+
+module.exports = transformer;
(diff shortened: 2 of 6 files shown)


sample/reverse-identifiers-async.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Example jscodeshift async transformer. Simply reverses the names of all
+ * identifiers.
+ */
+ function transformer(file, api) {
```
Daniel15, line 12: Make this `async` rather than returning a resolved promise? I think that'll more accurately reflect how people use it.

package.json
```diff
@@ -50,6 +50,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.1",
     "eslint": "^5.9.0",
+    "is-promise": "^4.0.0",
```
Daniel15, line 53: Do we really need to pull in a third-party library here?

src/testUtils.js
```diff
@@ -33,19 +34,42 @@ function applyTransform(module, options, input, testOptions = {}) {
...
 }
 exports.applyTransform = applyTransform;
 
 function runSnapshotTest(module, options, input) {
   const output = applyTransform(module, options, input);
+
+  if (isPromise(output)) {
+    return output.then(result => {
+      expect(result).toMatchSnapshot();
+
```
Daniel15, line 53: Nit: remove empty line

src/testUtils.js
```diff
@@ -33,19 +34,42 @@ function applyTransform(module, options, input, testOptions = {}) {
...
+  }
+
   return (output || '').trim();
 }
 exports.applyTransform = applyTransform;
 
 function runSnapshotTest(module, options, input) {
   const output = applyTransform(module, options, input);
+
+  if (isPromise(output)) {
```
Daniel15, line 50: Too bad we can't use async here, because then this could be a bit cleaner:

```
let output = applyTransform(.......);
if (isPromise(output)) {
  output = await output;
}
expect(output).toMatchSnapshot();
```

package.json
shiyangzhaoa: I want use `util.types.isPromise`, but it added in v10...

src/testUtils.js
shiyangzhaoa: Yes, I think so. On the other hand, using async/await will cause the return value to become a promise, I'm not sure if this will affect the existing logic...
I think these modifications are ugly, i'll close the PR.

PR #530, opened 2022-11-12
chore(github-actions): drop node12 support node18

chimurai:
- drop node12 support from github-actions (EOL since 2022-04-30)
- add support for node18


Daniel15:
Thanks!

PR #549, opened 2023-03-08
feat(crlf): add ending space to avoid env: node\r: No such file or di…

### LINKED ISSUE #424: Version: 0.12.0. /usr/bin/env: ‘node\r’: No such file or directory
Execute jscodeshift with an error
```
/usr/bin/env: ‘node\r’: No such file or directory 
```

sibelius:
…rectory, fix #424

### DIFF
--- a/bin/jscodeshift.js
+++ b/bin/jscodeshift.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node  
 
 /**
  * Copyright (c) Facebook, Inc. and its affiliates.


EmanuelCampos (commented):
Nice catch!

jakeboone02:
Hello, I just found this issue when trying to get `jscodeshift` to run with [Bun](https://bun.sh/). See oven-sh/bun#2516.

While I think Bun should ultimately work around the issue, the solution in this PR does not actually resolve it in Bun's case (see https://github.com/oven-sh/bun/issues/2516#issuecomment-1493140136), it just leads to other issues.

Is there any way `jscodeshift` could just ship LF line endings for `/bin/jscodeshift.js`? I think a `.gitattributes` file like this would work:

```
bin/jscodeshift.js eol=lf
```

I'd be happy to submit a PR to that effect.

Issue #561, opened 2023-06-28
extensions cli option not respected when passing individual file paths


robcmills:
## Description

If you run `jscodeshift` on the cli, with `--extensions=ts` provided, and you pass individual file paths (not a directory) like this:

```sh
jscodeshift -t transform.ts --extensions=ts readme.md
```

Then the extensions are _not respected_ and files that do not match those extensions are sent to the workers and processed through your transform file. Clearly, if the file is not the expected extension, then when your transform function tries to parse it, you will get an error:

<img width="1890" alt="Screenshot 2023-06-28 at 10 45 40 AM" src="https://github.com/facebook/jscodeshift/assets/4131746/632552ee-7869-4ddc-8575-e9a721057cbd">

I think this is because of this code right here in `jscodeshift/src/Runner.js`:

<img width="970" alt="Screenshot 2023-06-28 at 10 50 02 AM" src="https://github.com/facebook/jscodeshift/assets/4131746/96adb676-f48f-4ba1-a2dd-5e6724fa8056">

In this function, `filter` is passed in from this methods only usage:

<img width="882" alt="Screenshot 2023-06-28 at 10 55 43 AM" src="https://github.com/facebook/jscodeshift/assets/4131746/9ab8c12e-9005-404c-8747-40822b0fba41">

and you can see that it takes into account the extensions, filtering out any files that do not match. And in `getAllFiles`, you can see that _if the file is a directory, then it filters the files, but the else if condition does not._ The else if clause _only accounts for ignore patterns._ 

## **To Reproduce**

1. Run jscodeshift from the cli with `--extensions=ts` and pass it a filepath with a different extension

Note that it will not ignore that file and try to process it 
```
Processing 1 files...
```

## **Expected Behavior**

`--extensions` option should filter which files are processed, ignoring non-matching files

## **System Information:**

 - OS: Apple M1 Pro Ventura 13.4
 - Terminal: Warp v0.2023.06.20.08.04.stable_03
 - jscodeshift version: 0.15.0

## **Additional Context**

My use case, which I believe is quite common, is running jscodeshift as a part of a Github workflow automation, where we run a git command to get a list of files that are changed in a PR as compared to a development (or main) branch, and then pass those files to the codemod. This results in many individual filepaths (instead of a directory) that can include non-typescript files. 

[...]

PR #562, opened 2023-06-28
Respect extensions cli option when filtering individual files

robcmills:
Resolves https://github.com/facebook/jscodeshift/issues/561

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -148,7 +148,7 @@ function getAllFiles(paths, filter) {
             file,
             list => resolve(list.filter(filter))
           );
-        } else if (ignores.shouldIgnore(file)) {
+        } else if (!filter(file) || ignores.shouldIgnore(file)) {
           // ignoring the file
           resolve([]);
         } else {


Daniel15:
This could go either way - it's not defined whether passing in an exact filename should bypass the extension filter or not. Changing it so everything observes the extension filter seems reasonable to me though.

lg-kialo:
> This could go either way - it's not defined whether passing in an exact filename should bypass the extension filter or not. Changing it so everything observes the extension filter seems reasonable to me though.

@Daniel15 I disagree, for a single filename supplying the extension twice (once in the filename and once as `--extension` flag) seems redundant.

Daniel15:
@lg-kialo The `--extension` flag is optional. You don't have to provide it.

lg-kialo:
@Daniel15 you do have to provide it if you want to transform files with extension different than `.js`.
Before version 0.15.1 `jscodeshift -t transform.ts fileToTransform.tsx` used to work, now one is forced to use `jscodeshift -t transform.ts --extensions=tsx fileToTransform.tsx`

lg-kialo:
My suggestion would be that if no `--extensions` flag is passed, we don't filter by extension at all. Instead of filtering only files with the default extension (`.js`) as it is now.

PR #568, opened 2023-09-18
fix(worker): Update Worker.js

### LINKED ISSUE #560: how can I add @babel/plugin-proposal-private-methods to jscodeshift ?
node_modules/write-file-atomic/node_modules/signal-exit/dist/cjs/index.js: Class private methods are not enabled. Please add `@babel/plugin-proposal-private-methods` to your configuration.
  210 |         this.#emitter.count -= 1;

sibelius:
fix #560


sibelius:
why worker babel config is not the same as other babel config parser ?

Daniel15:
Seems reasonable to me.



> why worker babel config is not the same as other babel config parser ?

I'm not sure. I wonder if they should be unified.

PR #575, opened 2023-10-16
fix(unit tests): Fix for unit tests broken by #562

ElonVolo:
Fixing unit tests that were broken because the unit test framework by default looks for an extension with .js and the tempfiles created did not have a .js extension. Which wasn't problem, until #562 finally got the single file filter functionality working the way it was supposed to all along.


Daniel15:
Thank you so much!

PR #585, opened 2024-02-26
Include importAttributes in tsOptions.js

benasher44:
Allow syntax like `import obj from "./something.json" with { type: "json" };`


Daniel15:
Released in v17.0.0. Thanks for your contribution!

Issue #580, opened 2024-01-05
Enable parsing of `accessor` keyword


RhodeHummel:
It looks like jscodeshift does not currently support the accessor keyword that was added typescript in 3.9. Can we please add support for this. Thank you.

https://github.com/microsoft/TypeScript/pull/49705
https://github.com/tc39/proposal-decorators

ElonVolo:
Would you be able to add the content you were trying to transform, the transformation script you wrote, the command you ran to transform it, and the expected output, to the issue?

Daniel15:
I think this will need upstream changes (recast)

RhodeHummel:
@ElonVolo Here is a simple case. Basically, I just want it to run with out any errors. Hope this is enough information for you.

```ts
// test-file.ts
export class Test {
    public accessor myValue = 10;
}
```

```ts
// test-transform.ts
import { FileInfo, API, Options } from 'jscodeshift';

module.exports = function (file: FileInfo, { j }: API, _options: Options) {
    const root = j(file.source);

    // Do nothing just output the file again
    return root.toSource();
};
```

ElonVolo:
I'm running the tranform and file code you submitted and I'm not getting any errors. Nothing's being transformed, which is expected as no modifications have been made.

Are you getting some different result on your end?

ElonVolo:
> I think this will need upstream changes (recast and/or ast-types)

I looked at the recast code and there does seem to be some support for Typescript's **accessor** keyword.

RhodeHummel:
@ElonVolo I definitely get an error. There is no error message, it just prints out results 1 error instead of 1 unmodified. I have created an isolated test case and included it here.

To see the error, run `npx ts-node test.ts`

[jscodeshift-error.zip](https://github.com/facebook/jscodeshift/files/13906755/jscodeshift-error.zip)

PR #594, opened 2024-07-10
fix: add decoratorAutoAccessors plugin

syi0808:
add decoratorAutoAccessors plugin in typescript parser plugins to solve error.

### before screenshot
Screenshot 2024-07-10 at 9 02 52 PM

### after screenshot
Screenshot 2024-07-10 at 9 02 41 PM

fix: #580


Daniel15:
Can you please provide an example of code that requires this option?

ElonVolo:
Here's an example, @Daniel15 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html

To kill two birds with one stone, I think it would be great if in addition to providing the awesome fix for this bug, syi0808 could also add a unit test(s) that make use of the accessor keyword in some example code and then makes sure the manipulation of it comes out correctly.

That way we can guard against any future regressions with the parsing.

syi0808:
Thanks for reviews! I add test for changes now. But, before this pr e2e test didn't exist. So i add separated test for e2e. Will it be okay?

doouding:
I encountered this problem too. Looking forward to it.

test/__tests__/ts-decorator-auto-accessor-test.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
```
Daniel15, line 1: Can these go in the existing `src/__tests__` directory?

test/__tests__/ts-decorator-auto-accessor-test.js
```diff
@@ -0,0 +1,10 @@
...
+
+jest.autoMockOff();
+const defineTest = require("../../src/testUtils").defineTest;
+
+describe("should be parse typescript decoratorAutoAccessors correctly", function () {
+  defineTest(__dirname, "ts-decorator-auto-accessor", null, null, {
```
Daniel15, line 7: Please use `defineInlineTest`. You can also define the transform inline if you use `defineInlineTest`, e.g.

```
function transformer(file, api) {
  .....
}

defineInlineTest(transformer, .......)
```

Daniel15 (approved):

Daniel15:
Sorry for the delay in getting back to this. I added some comments inline.

test/__tests__/ts-decorator-auto-accessor-test.js
syi0808: Okay. I moved the test.

test/__tests__/ts-decorator-auto-accessor-test.js
syi0808: Thanks for review. I switched to inline test.

Daniel15:
Thanks!

Daniel15:
Released in v17.0.0. Thanks for your contribution!

PR #595, opened 2024-07-12
docs: update getVariableDeclarators description

morinokami:
The [current description](https://jscodeshift.com/build/api-reference/#getvariabledeclarators) of getVariableDeclarators seems to be inaccurate (it says it takes a string as an argument, like findVariableDeclarators, but the actual argument is a callback), so I fixed it.

See also: https://github.com/facebook/jscodeshift/blob/main/src/collections/Node.js#L103


Daniel15:
Thank you!

PR #596, opened 2024-07-12
docs: add description for `nodes` api

morinokami:
Added missing explanation of the [`nodes`](https://github.com/facebook/jscodeshift/blob/main/src/Collection.js#L158) method.


Daniel15:
Thank you :)

Issue #598, opened 2024-07-15
Use corepack for managing yarn version


trivikr:
### Is your feature request related to a problem? Please describe.

The jscodeshift project uses yarn for development, but doesn't manage yarn version using [corepack](https://yarnpkg.com/corepack) recommended by yarn team.

```console
$ node -v
v20.15.1

$ yarn
zsh: command not found: yarn

$ corepack -v
0.28.1

$ corepack enable

$ yarn
! The local project doesn't define a 'packageManager' field. Corepack will now add one referencing yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447.
! For more details about this field, consult the documentation at https://nodejs.org/api/packages.html#packagemanager

yarn install v1.22.19
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...

warning Your current version of Yarn is out of date. The latest version is "1.22.22", while you're on "1.22.19".
info To upgrade, run the following command:
$ curl --compressed -o- -L https://yarnpkg.com/install.sh | bash
$ cp -R src/ dist/
✨  Done in 2.70s.
```

### Describe the solution you'd like
Use corepack for managing yarn version

PR #599, opened 2024-07-15
Use corepack for managing yarn version

trivikr:
Fixes: https://github.com/facebook/jscodeshift/issues/598

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,12 +12,13 @@ the process for contributing will follow standard GitHub protocol.
 ## Pull Requests
 We actively welcome your pull requests.
 
-1. Fork the repo and create your branch from `master`.
-2. If you've added code that should be tested, add tests.
-3. If you've changed APIs, update the documentation.
-4. Ensure the test suite passes.
-5. Make sure your code lints.
-6. If you haven't already, complete the Contributor License Agreement ("CLA").
+1. Fork the repo and create your branch from `main`.
+1. Run `corepack enable` to set up `yarn` package manager.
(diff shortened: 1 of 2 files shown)


Daniel15:
The Node.js site says that Corepack is experimental: https://nodejs.org/api/corepack.html. Do you know if there's a risk of breaking changes that could cause issues, or a timeline for a stable release?

trivikr:
> Do you know if there's a risk of breaking changes that could cause issues

If Node.js removes corepack, most developers on yarn will ideally switch to installing it globally using `npm install -g corepack`, and it'll work in the same way.

> a timeline for a stable release?

No one knows. We've been trying to push since May'22 in https://github.com/nodejs/corepack/issues/104, but Node.js TSC is not able to arrive at a decision.

> The readme should probably also have non-Corepack instructions for people using older Node.js versions.

Corepack is available in all Node.js versions supported by jscodeshift. And these instructions are for contributors of jscodeshift, and not users.

Daniel15:
Sounds good to me! Thank you.

Daniel15:
@trivikr  We should probably add an SHA hash as documented in the Corepack readme: https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages to reduce the risk of supply chain attacks (having said that, I've been doing the Yarn 1.x packaging, and I trust myself :stuck_out_tongue:)

trivikr:
> We should probably add an SHA hash as documented in the Corepack readme: https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages

PR to add sha https://github.com/facebook/jscodeshift/pull/602

Issue #600, opened 2024-07-15
Replace `chalk` dependency with a lighter alternative


trivikr:
### Is your feature request related to a problem? Please describe.

Recently, the [Ecosystem Performance](https://e18e.dev/) was formed to cleanup, speedup, levelup the JavaScript Ecosystem.
Socket Security team wrote a blog post about it https://socket.dev/blog/javascript-community-launches-e18e-initiative-to-improve-ecosystem-performance

On the e18e cleanup channel on Discord, multiple solutions are being discussed. One of them is to replace `chalk` with `picocolors`. I see that jscodeshift uses `chalk` https://github.com/facebook/jscodeshift/blob/0994ad8361a37e6f7c4f167c3ddfc73ebb2083f3/package.json#L37

### Describe the solution you'd like
Replace `chalk` with a lighter alternative, like `picocolors`.

trivikr:
I created a similar feature request with jscodeshift devDependencies which uses chalk:
* jest https://github.com/jestjs/jest/issues/15189
* eslint https://github.com/eslint/eslint/issues/18686

Babel already replaced `chalk` with `picocolors` in https://github.com/babel/babel/pull/16359

Daniel15:
What does "light" mean in this case? Could you please share the metrics being used?

trivikr:
> What does "lighter" mean in this case? Could you please share the metrics being used?

It usually means an alternative with a smaller size (install/publish/bundle).

The dependency `chalk@4.x` which jscodeshift uses has [~33 kB publish size and ~105 kB install size](https://packagephobia.com/result?p=chalk@4.0.0).

<details>
<summary>Screenshot</summary>

<img width="767" alt="Screenshot 2024-07-15 at 8 32 55 AM" src="https://github.com/user-attachments/assets/8351a140-67d6-4414-b8a8-27dd900b4de9">

</details>

On other hand, picocolors has [5.03 kB install/publish size](https://packagephobia.com/result?p=picocolors)

<details>
<summary>Screenshot</summary>

<img width="588" alt="Screenshot 2024-07-15 at 8 34 10 AM" src="https://github.com/user-attachments/assets/4bad253b-2eb1-4c7f-b067-3f2e1142d554">

</details>

The lighter alternatives sometimes come with performance benefits too.
The picocolors benchmarks state it's 2X faster than chalk in their README.

PR #601, opened 2024-07-15
Migrate from chalk to picocolors

trivikr:
Fixes: https://github.com/facebook/jscodeshift/issues/600

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1130,7 +1130,7 @@ chalk@^2.4.2:
     escape-string-regexp "^1.0.5"
     supports-color "^5.3.0"
 
-chalk@^4.0.0, chalk@^4.1.2:
+chalk@^4.0.0:
   version "4.1.2"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -1130,7 +1130,7 @@ chalk@^2.4.2:
     escape-string-regexp "^1.0.5"
     supports-color "^5.3.0"
 
-chalk@^4.0.0, chalk@^4.1.2:
+chalk@^4.0.0:
```
Daniel15, line 1133: What else is still pulling `chalk` in?

trivikr: Chalk is used in devDependencies of jscodeshift, like jest and eslint.

I've created feature requests with those repos, documented as a follow-up in https://github.com/facebook/jscodeshift/issues/600#issuecomment-2227665223

Daniel15:
Thanks :)

PR #602, opened 2024-07-15
Add SHA hash of yarn@1.22.22 in packageManager field

trivikr:
Refs: https://github.com/facebook/jscodeshift/pull/599#issuecomment-2229122985


Daniel15:
Very nice

Issue #606, opened 2024-07-16
Define Node.js supported version in the package.json engines field


trivikr:
### Is your feature request related to a problem? Please describe.

The supported Node.js version is not called out in an easy to find place in jscodeshift project.

For example, when upgrading to 0.16.0, I'd to read through PR https://github.com/facebook/jscodeshift/pull/588 which mentioned in the description
> To support the updates to these libraries it will be necessary to deprecate node 12 and some node 14 versions prohibited by the libraries.

There are several ways to call out supported Node.js versions on npm packages. The official one is to mention it in the [engines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) field.

### Describe the solution you'd like
Define Node.js supported version in the package.json engines field as `>=16.0.0`

### Describe alternatives you've considered
N/A

PR #607, opened 2024-07-16
Specify node >=16 as supported in package.json engines field

trivikr:
Fixes: https://github.com/facebook/jscodeshift/issues/606


Daniel15 (approved):
Looks good to me.

PR #618, opened 2024-08-06
Replace `temp` package with vendored code to avoid deprecated `rimraf` transitive dependency

kitten:
This is simply to avoid the dependency chain on `rimraf` via `temp@0.8.4 -> rimraf@2.6.3`, which has been deprecated. The [`temp` package](https://github.com/bruce/node-temp) hasn't seen a release in a while, and minimal functionality is used from it. The `temp.track` functionality also doesn't seem to be used, which would make it pretty trivial to vendor the code.

Switching to alternatives like `tempy` likely isn't attractive either since it's ESM-only while `jscodeshift` is not. Older versions of `tempy` also depend transitively on deprecated versions of `rimraf`.

Tests pass and cover changes as-is.

**Note:** The vendored code has an added copyright notice line for `node-temp`.

### DIFF
--- a/src/utils/temp.js
+++ b/src/utils/temp.js
@@ -34,17 +34,11 @@ function generateName(suffix) {
   );
 }
 
-exports.open = function open(opts, callback) {
+exports.open = async function open(opts) {
   if (!opts) opts = {};
   const target = generateName(opts.suffix);
-  return new Promise((resolve, reject) => {
-    fs.open(target, RDWR_EXCL, 0o600, (error, fd) => {
-      const result = { path: target, fd };
-      if (callback) callback(error, result);
-      if (error) return reject(error);
-      resolve(result);
-    });
-  });
+  const fd = await fs.promises.open(target, RDWR_EXCL, 0o600);
+  return { path: target, fd };
 };
 
 exports.openSync = function open(opts) {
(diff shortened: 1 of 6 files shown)


src/utils/temp.js
```diff
@@ -0,0 +1,58 @@
...
+      (Math.random() * 0x100000000 + 1).toString(36),
+      suffix || '',
+    ].join('')
+  );
+}
+
+exports.open = function open(opts, callback) {
+  if (!opts) opts = {};
+  const target = generateName(opts.suffix);
+  return new Promise((resolve, reject) => {
```
Daniel15, line 40: Can this use async/await with a promisified version of `fs.open` given the minimum Node version is v16 now?

kitten: totally! I'll get that updated 👍 I wasn't sure about whether I should match existing patterns or not

kitten: @Daniel15: That's moved over to `fs.promises`. I didn't refactor `Runner.js` fully as I didn't want to make this PR overstep what it's supposed to do.
I also caught some invalid code that wasn't cleaned up properly.

kitten: I actually just noticed that the existing tests aren't covering this and haven't been failing despite the problematic code that I forgot to clean up, I'll have a look at what happened there

kitten:
@Daniel15: Sorry about that! ❤️ There's a test in the latest commit now that checks that the remote transformer files actually work.

trivikr:
The `temp` was replaced by `tmp` in https://github.com/facebook/jscodeshift/pull/633

PR #622, opened 2024-08-29
feat: add pkg.pr.new

Aslemammad:
This PR adds continuous preview releases to jscodeshift! 

Before running the workflow, we'd have to install the [app](https://github.com/apps/pkg-pr-new)!

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -1,0 +1,21 @@
+name: Continuous Releases
+
+on:
+  push:
+  pull_request:
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.x
+          cache: yarn
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+      - name: Preview
+        run: npx pkg-pr-new publish --compact
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -18,4 +18,4 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile
       - name: Preview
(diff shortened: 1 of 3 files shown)


Daniel15:
This sounds useful. Do you have a demo of how it works?

.github/workflows/cr.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.x
+          cache: yarn
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+      - name: Preview
+        run: npx pkg-pr-new publish --compact
```
Daniel15, line 21: Is there a way to lock this to a particular version or hash of `pkg-pr-new` so that we have to manually bump it to new versions? Running arbitrary new versions of this in a build script is a major supply chain attack risk.

Daniel15 (changes requested):
See inline comment

Aslemammad:
Amazing, I just moved to yarn now and installed pkg.pr.new! Would be happy to hear your thoughts!

Daniel15:
Sure, let's try it out!

Daniel15:
> Before running the workflow, we'd have to install the [app](https://github.com/apps/pkg-pr-new)!

I sent a request for access (to the admins of the Facebook org on Github, I guess) since I don't have permission to do this myself.

Daniel15:
I just noticed that this breaks our Node 16.x and 18.x builds since some of the dependencies depend on 20.x. Sorry, I should have checked that before I merged this. Is there some way around that?

Daniel15:
@Aslemammad  Do you know what this error means?

> There is no workflow defined for nFAlv21cKx

Aslemammad:
I see, have you installed the github application?

Daniel15:
Looks like this is working now. I had to get some issues fixed with the GitHub app. https://github.com/facebook/jscodeshift/pull/637

PR #629, opened 2024-09-17
Revert pkg-pr-new to using npx

Daniel15:
since it breaks older Node versions when it's in package.json.


Daniel15:
The breakage is because the app is not installed yet. I'm waiting on permission to install it.

alexbit-codemod:
fyi @Aslemammad

PR #631, opened 2024-09-20
Fix renameTo false positive for JSX tags

henryqdineen:
This fixes an edge case that we ran into where JSX tags are inadvertently renamed when they are "built-in" components like `<div>` or `<span>`. We should only consider capitalized tags as variable references (see [React docs](https://legacy.reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized)). For this PR I used the the [same regex](https://github.com/babel/babel/blame/368b512229d41147f5ea835a3dd7d5a256a5764e/packages/babel-types/src/validators/react/isCompatTag.ts#L3) as the JSX Babel transform for consistency. 

I understand that this logic is React-specific and there could be JSX usage that does not follow these conventions. Ideally this could be configurable.

### DIFF
--- a/src/collections/VariableDeclarator.js
+++ b/src/collections/VariableDeclarator.js
@@ -157,6 +157,15 @@ const transformMethods = {
             return false;
           }
 
+          if (
+            (types.JSXOpeningElement.check(parent) || types.JSXClosingElement.check(parent)) &&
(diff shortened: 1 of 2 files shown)


Daniel15:
Good catch! Thanks.

PR #633, opened 2024-09-27
chore: replace `temp` library with `tmp`

r4zendev:
This PR would fix the deprecation warnings that appear during the installation of `jscodeshift`'s peer dependencies.
An example of such warnings can be reproduced by installing the `jscodeshift` CLI tool globally (`npm i -g jscodeshift`) or by installing any other package that has jscodeshift as a dependency. For instance, this is what the error looks like in my shell:
Screenshot 2024-09-27 at 23 16 00

The testing of this PR was done by performing the following steps:
- Packing the app with updated dependencies into a tarball using `yarn pack`
- Installing it globally via `npm i -g ./path/to/tarball.tgz`

The results are as shown:
Screenshot 2024-09-27 at 23 16 08

After this PR is merged, a release will be required to fix the error for all of the dependants.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "neo-async": "^2.5.0",
     "picocolors": "^1.0.1",
     "recast": "^0.23.9",
-    "temp": "^0.9.4",
+    "tmp": "^0.2.3",
     "write-file-atomic": "^5.0.1"
   },
   "peerDependencies": {
--- a/bin/__tests__/jscodeshift-test.js
+++ b/bin/__tests__/jscodeshift-test.js
@@ -17,9 +17,7 @@ jest.setTimeout(600000); // 10 minutes
 const child_process = require('child_process');
 const fs = require('fs');
 const path = require('path');
-const temp = require('temp');
 const testUtils = require('../../utils/testUtils');
-const {chdir} = require('process');
 
 const createTransformWith = testUtils.createTransformWith;
 const createTempFileWith = testUtils.createTempFileWith;
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -14,8 +14,10 @@ const fs = require('graceful-fs');
 const path = require('path');
 const http = require('http');
 const https = require('https');
-const tmp = require('tmp');
 const ignores = require('./ignoreFiles');
+
+const tmp = require('tmp');
+tmp.setGracefulCleanup();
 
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -54,9 +54,11 @@
     "@babel/eslint-parser": "^7.24.7",
     "eslint": "8.56.0",
     "jest": "^29.7.0",
+    "jest-environment-node-single-context": "^29.4.0",
```
Daniel15, line 57: What does this do?

Daniel15 (commented):
Seems reasonable to me.

package.json
r4zendev: This mitigates the error that appears when running `yarn test`. The `fs-temp`'s utility chain of imports leads to a package named `to-data-view` which performs the following check:
image

Somehow, the data gets transformed into a `Uint8Array` and gets passed to this function, where it yields false for these checks ONLY in `jest`. Spending some time on Github issues, I've found a suggestion that this is caused by `jest` and `node` having different contexts during the test suite run. This testing environment dependency helped to solve this issue.
I've tried testing the same test with the same function that uses `fs-temp` to create temporary files in a testing project of my own and it was not happening, but in `jscodeshift` codebase it somehow did. So, instead of trying to update `jest` and/or write a custom environment file, I used this existing package.

r4zendev: You know what, I removed the `fs-temp` dependency too. Let me know if the new implementation needs improvements.

Daniel15:
Thanks! This is looking pretty good. I like 0-dependency libraries. Looked through the code of the module and it seems reasonable to me.

Can you please call the `setGracefulCleanup` method somewhere so that the library automatically deletes the temp files once the process exits?

Daniel15:
Looks good. Thanks!

trivikr:
When is this change expected to be published? Will it be in `v17.0.1` or `v17.1.0`?

Daniel15:
Released as v17.1.0. I usually wait for more substantive changes to bundle them all together into one release, to avoid a lot of churn. People don't like seeing updates all the time. That's why it took a while to release. We didn't end up with any other major changes in this release.

trivikr:
Looks like there was some issue with publishing v17.1.0 which includes removed dependency `v17.1.0`

I've created a bug report at https://github.com/facebook/jscodeshift/issues/638

Issue #454, opened 2021-09-24
defineInlineTest doesn't support async transforms


robatwilliams:
Since #237, transformers can be async.

However testing such transformers using `defineInlineTest` doesn't appear to be supported, and from the code also looks like it won't be suppored by any type of unit test.

Error: `TypeError: (output || "").trim is not a function`

MonicaOlejniczak:
I have the same problem :(

danieldelcore:
Hey all, in case anyone finds it useful here's an alternate testing util that supports async transforms. Feel free to either copy-paste or pull the library as a dep. 

The API is different to `defineInlineTest` because it allows for you to still have access to jest library features like `it.skip`, `it.only` etc.

## [applyTransform](https://www.codeshiftcommunity.com/docs/test-utils/#applytransformtransform-input-options---parser-babel-)

**Example:**

```ts
import * as transformer from '../transform';
import { applyTransform } from '@codeshift/test-utils';

it('should wrap avatar in a tooltip if name is defined', async () => {
  const result = await applyTransform(transformer, `const foo = 'foo';`);

  expect(result).toMatchInlineSnapshot(`"const oof = 'oof';"`);
});
```

here's the implementation in case you would rather copy-paste:

```ts
import jscodeshift from 'jscodeshift';

type Parser = 'babel' | 'babylon' | 'flow' | 'ts' | 'tsx';

interface Options {
  parser?: Parser;
}

export default async function applyTransform(
  transform: any,
  input: string,
  options: Options = {
    parser: 'babel',
  },
) {
  // Handle ES6 modules using default export for the transform
  const transformer = transform.default ? transform.default : transform;
  const output = await transformer(
    { source: input },
    {
      jscodeshift: jscodeshift.withParser(options.parser as string),
      stats: () => {},
    },
    options || {},
  );

  return (output || '').trim();
}
```

PR #646, opened 2024-12-21
Enable async tranformers in test utils

PaquitoSoft:
Currently the available [test utilities](https://github.com/facebook/jscodeshift?tab=readme-ov-file#unit-testing) exposed by the library helps a lot when developers need to test their custom transformation but they only support synchronous transformations.

I've seen others also asking about this feature ([reference1](https://github.com/facebook/jscodeshift/issues/516), [reference 2](https://github.com/facebook/jscodeshift/issues/454)) so I've decided to give it a try and update those utilities to be able to manage both sync and async transformations.

Please let me know if this is something that could potentially be adopted by `jscodeshift`.

Thanks.

### DIFF
--- a/.changeset/polite-lizards-know.md
+++ b/.changeset/polite-lizards-know.md
@@ -1,0 +1,5 @@
+---
+"jscodeshift": patch
+---
+
+Enable async tranformers in test utils.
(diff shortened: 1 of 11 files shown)


Daniel15:
Thanks for this! 

Can you please add a changeset as per the second comment? We'retrying to start using [changesets](https://github.com/changesets/changesets) to manage versioning/releases.

PaquitoSoft:
@Daniel15 Sorry I forgot about it.

I've just added it in this commit: 71f14399d879eb1cb767834f5aed39b3e259829b

Is it enough?

Daniel15:
Thanks. I just got back from vacation. I'll review this soon. I need to learn about changesets myself :)

PR #666, opened 2025-05-19
fix(readme): passing `recast` options

mbtools:
- Fix link to `recast` config options
- Add example for using `lineTerminator` which is a common issue

### DIFF
--- a/.changeset/small-kangaroos-beg.md
+++ b/.changeset/small-kangaroos-beg.md
@@ -1,0 +1,5 @@
+---
+"jscodeshift": patch
+---
+
+fix(readme): passing `recast` options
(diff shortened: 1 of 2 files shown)


README.md
ElonVolo, line 1: Looks good to me.

.changeset/small-kangaroos-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"jscodeshift": patch
+---
+
+fix(readme): passing `recast` options
```
Daniel15, line 5: We don't need changesets for the readme, since it doesn't affect the released code. You can remove this and just keep the readme changes :)

mbtools: The changeset bot confused me. Maybe you tweak it a bit:

image

PR #680, opened 2026-01-07
Add iterator implementation to Collections

codingjoe:
ES6 iterators support the new for..of syntax. Since a `Collection` wraps an `Array`, we can lean on ECMAScript's new iterator delegation.

Resolve #679

I am working on [esupgrade](https://github.com/codingjoe/esupgrade) a tool to upgrade JS/TS code to widely available ECMAScript features. It's built on this wonderful tool. I'd really love to keep esupgrade's code base up to date too, and using for..of loops and generators would help tremendously.

Thanks for your consideration and for publishing this under a free license <3

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -65,6 +65,15 @@ class Collection {
   }
 
   /**
+   * Implements for-of iterator semantic for node paths of the collection.
+   *
+   * @returns {Iterator<import("ast-types").NodePath>}
+   */
+  *[Symbol.iterator]() {
+    yield* this.__paths;
+  }
+
+  /**
    * Executes callback for each node/path in the collection.
    *
    * @param {function} callback
--- a/.changeset/metal-walls-switch.md
+++ b/.changeset/metal-walls-switch.md
@@ -1,0 +1,7 @@
+---
+"jscodeshift": minor
+---
+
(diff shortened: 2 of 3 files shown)


src/Collection.js
```diff
@@ -64,6 +64,15 @@ class Collection {
...
 
+  /**
+   * Implements for-of iterator semantic for node paths of the collection.
+   *
+   * @returns {Iterator<import("ast-types").NodePath>}
+   */
+  *[Symbol.iterator]() {
+    yield* this.__paths;
+  }
+
```
Daniel15, line 75: I wonder if we should have a nodes iterator too, like the existing `nodes` method. It could be called `iterateNodes` or something similar?

Daniel15:
Looks pretty good to me! Just left a small comment inline.

src/Collection.js
codingjoe: I looked into the method. Since it returns a native Array, it is already iterable.
e.g., the following code will already work:

```javascript
for (const node of colletion.nodes()) { console.debug(node) }
```

If you like, I can add a regression test for this behavior, but maybe it's a little extra.

codingjoe:
<3 thx

PR #694, opened 2026-04-02
Security: Remote transform URLs can lead to arbitrary code execution

tuanaiseo:
## Problem

The CLI accepts `--transform` values that match `http(s)` and forwards them to `Runner.run(...)`. Since transforms are executable JavaScript loaded and run by workers, allowing remote URLs creates an explicit remote code execution path if the URL is untrusted or can be tampered with (MITM, compromised host, internal redirect abuse).

**Severity**: `high`
**File**: `bin/jscodeshift.js`

## Solution

Disable remote URL transforms by default. Add an explicit opt-in flag (for example `--allow-remote-transform`) plus allowlist validation (trusted hosts, HTTPS-only, optional checksum/signature verification) before executing fetched code.

## Changes

- `bin/jscodeshift.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed

### DIFF
--- a/bin/jscodeshift.js
+++ b/bin/jscodeshift.js
@@ -26,9 +26,16 @@ const parser = require('../src/argsParser')
       display_index: 15,
       abbr: 't',
       default: './transform.js',
-      help: 'path to the transform file. Can be either a local path or url',
+      help: 'path to the transform file. Can be either a local path or url (requires --allow-remote-transform)',
       metavar: 'FILE',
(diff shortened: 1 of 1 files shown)


Daniel15:
I'm sorry but I don't understand the point of this extra command-line argument. If the user doesn't want to execute remote code, they shouldn't pass a URL to begin with.

PR #695, opened 2026-04-02
Security: Automatic Babel transpilation executes untrusted transform code in-process

tuanaiseo:
## Problem

When Babel mode is enabled, the worker uses `@babel/register` and then `require(tr)` to load and execute the transform module. In multi-tenant or service deployments, a user-provided transform path results in arbitrary code execution with worker privileges, including filesystem and network access.

**Severity**: `high`
**File**: `src/Worker.js`

## Solution

Treat transforms as untrusted code: run in a hardened sandbox (separate container/user, seccomp/AppArmor, read-only FS where possible, blocked outbound network), or require signed/approved transforms. Consider disabling Babel/register for untrusted scenarios.

## Changes

- `src/Worker.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed

### DIFF
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -29,6 +29,8 @@ let notify;
 let transform;
 let parserFromTransform;
 
+const allowUnsafeBabel = process.env.JSCODESHIFT_ALLOW_UNSAFE_BABEL === '1';
+
 if (module.parent) {
   emitter = new EventEmitter();
   emitter.send = (data) => { run(data); };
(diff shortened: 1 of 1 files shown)


Daniel15:
I'm not sure this should be done in jscodeshift itself. The jscodeshift user could just directly execute the code themselves.


If a shared environment doesn't want this feature to be used, it shouldn't expose the feature to users.

PR #705, opened 2026-04-11
fix: resolve #420 — Missing parentheses around LogicalExpression inside of UnaryExpression

### LINKED ISSUE #420: Missing parentheses around LogicalExpression inside of UnaryExpression
I believe I've stumbled onto a bug in jscodeshift: I created a transform to convert
```
(a && b) || c();
```
to
```
if (!(a && b)) {
  c();
}
```

However, when I wrap the initial `LogicalExpression` in a `!` `UnaryExpression`, `jscodeshift` instead outputs:
```
if (!a && b) {
  c();
}
```
which has a different meaning due to the missing parentheses.

The issue on AST explorer: https://astexplorer.net/#/gist/4f8d333d868a270bd53098d4f283e8fa/5cb2224e8c7cbc178130c3da2ab6683e4a1064bc

Output of `jscodeshift --version`:
```
jscodeshift: 0.11.0
 - babel: 7.13.14
 - babylon: 7.13.13
 - flow: 0.148.0
 - recast: 0.20.4
```

I don't believe this is an issue with recast, because I wrote some code that does the same thing purely using recast, and it works:
```
var recast = require("recast");
const b = recast.types.builders;

const code = [
  "(a && b) || c();"
].join("\n");

const ast = recast.parse(code);
const expression = ast.program.body[0].expression;
const body = b.blockStatement([b.expressionStatement(expression.right)]);
const cond = b.unaryExpression("!", expression.left);
const ifStatement = b.ifStatement(cond, body, null);
ast.program.body[0] = ifStatement;
const output = recast.print(ast).code;

console.log(output);
```
outputs:
```
if (!(a && b)) {
  c();
}
```

ed-thuando:
## Summary

fix: resolve #420 — Missing parentheses around LogicalExpression inside of UnaryExpression

## Problem

**Severity**: `High` | **File**: `src/core.js`

When jscodeshift transforms code like `(a && b) || c();` to `if (!(a && b)) { c(); }`, it outputs `if (!a && b) { c(); }` instead. This happens because when a LogicalExpression is wrapped in a UnaryExpression, the necessary parentheses are not being preserved. The operator `!` has higher precedence than `&&`, so `!a && b` means `( !a ) && b` rather than `!(a && b)`.

## Solution

Look at the `printOrReplace` function in core.js to see how it handles printing expressions. The issue is likely that when printing a UnaryExpression with a LogicalExpression as its argument, the code doesn't check if parentheses are needed based on operator precedence. The fix should check if the argument of a UnaryExpression needs parentheses (i.e., if it's a LogicalExpression or other lower-precedence expression) and add them accordingly. This is similar to how recast handles this, but jscodeshift may be using a custom print function that doesn't account for this case.

## Changes

- `src/core.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed

---

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -18,6 +18,32 @@ const template = require('./template');
 const Node = recast.types.namedTypes.Node;
 const NodePath = recast.types.NodePath;
 
+const namedTypes = recast.types.namedTypes;
+const builders = recast.types.builders;
+
+const Precedence = {
+  UNARY: 15,
+  LOGICAL_AND: 13,
+  LOGICAL_OR: 14,
+  SEQUENCE: 1,
+};
+
+function needsParens(path, childPath) {
+  const node = path.value;
+  const childNode = childPath.value;
+  
+  if (node.type === namedTypes.UnaryExpression.type) {
+    if (childNode.type === namedTypes.LogicalExpression.type) {
+      return true;
+    }
+    if (childNode.type === namedTypes.SequenceExpression.type) {
+      return true;
+    }
+  }
+  
+  return false;
+}
+
 // Register all built-in collections
 for (var name in collections) {
   collections[name].register();


src/core.js
```diff
@@ -18,6 +18,32 @@ const template = require('./template');
 const Node = recast.types.namedTypes.Node;
 const NodePath = recast.types.NodePath;
 
+const namedTypes = recast.types.namedTypes;
+const builders = recast.types.builders;
+
+const Precedence = {
```
Daniel15, line 24: This isn't used?

src/core.js
```diff
@@ -18,6 +18,32 @@ const template = require('./template');
...
+  UNARY: 15,
+  LOGICAL_AND: 13,
+  LOGICAL_OR: 14,
+  SEQUENCE: 1,
+};
+
+function needsParens(path, childPath) {
```
Daniel15, line 31: This isn't called from anywhere?

Daniel15 (changes requested):
The added code isn't actually called from anywhere. Please add tests too.

PR #725, opened 2026-07-14
Fix ignore patterns not matching relative source paths starting with ../ (#556)

### LINKED ISSUE #556: Ignore pattern doesn't work with relative paths starting with two dots (../)
I'm running my transform with command
`npx jscodeshift -t transforms/replace-indexes.ts ../company.reactapp/src --ignore-pattern **/node_modules/**`
I would expect the command to ignore all files in the `node_modules` directory. What happens instead is no files is ignored. 
Took me whole day to figure out that the relative path coused the issue. If I used an absolute path the the ignore pattern works just fine.

What little I managed to test, the true issue might be in `micromatch` library that handles the GLOB matching.

LeSingh1:
## Summary

Fixes #556.

`--ignore-pattern` (and every other ignore source that funnels through `shouldIgnore`) silently fails to match files when the source path is **relative and starts with `../`**. For example:

```sh
jscodeshift -t transform.ts ../app/src --ignore-pattern '**/node_modules/**'
```

matches nothing, so files under `node_modules` are transformed anyway, while the exact same invocation with an **absolute** source path (or a relative path without a leading `../`) ignores them correctly. Several users independently reported losing time to this on #556.

## Root cause

The directory walker in `Runner.js` builds candidate paths with `path.join(dir, file)`. When the source argument is `../app/src`, `path.join` preserves the leading `../`, so `shouldIgnore` receives paths like `../app/src/node_modules/foo.js`.

`shouldIgnore` matches these with picomatch, but picomatch's `**` wildcard does **not** match across a leading `../` (or `./`) traversal segment:

```js
const picomatch = require('picomatch');
picomatch.isMatch('src/node_modules/foo.js',        '**/node_modules/**', { dot: true }); // true
picomatch.isMatch('/abs/src/node_modules/foo.js',   '**/node_modules/**', { dot: true }); // true
picomatch.isMatch('../app/src/node_modules/foo.js', '**/node_modules/**', { dot: true }); // false  <-- the bug
```

## Fix

`shouldIgnore` now also tests the path with any leading `../`/`./` traversal prefix stripped, so a relative source path behaves the same as an absolute one. The original path is still matched first, so this is strictly additive — no path that was previously ignored stops being ignored (e.g. a literal `../secret.js` pattern still matches).

## Test plan

Added `src/__tests__/ignoreFiles-test.js` (the module had no direct coverage before). The new cases fail on `main` and pass with this change:

- glob and bare-directory patterns ignore matching `../`/`./`-prefixed paths
- absolute and plain-relative matching is unchanged
- non-matching `../` paths are still not ignored

```
Test Suites: 19 passed, 19 total
Tests:       1 skipped, 214 passed, 215 total
```

A `patch` changeset is included.

### DIFF
--- a/.changeset/ignore-relative-parent-paths.md
+++ b/.changeset/ignore-relative-parent-paths.md
@@ -1,0 +1,7 @@
+---
+"jscodeshift": patch
+---
+
+Fix `--ignore-pattern` not matching files when the source path is relative and starts with `../` (or `./`)
+
+Ignore patterns such as `**/node_modules/**` were silently skipped for files enumerated from a relative source path like `../app/src`, because picomatch's `**` wildcard does not match across a leading `../`/`./` traversal segment. The path is now also matched with any leading traversal prefix stripped, so ignore behavior is consistent for relative and absolute source paths.
--- a/src/__tests__/ignoreFiles-test.js
+++ b/src/__tests__/ignoreFiles-test.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/*global jest, describe, it, expect, beforeEach*/
+
+'use strict';
+
+describe('ignoreFiles', function() {
+  let ignores;
+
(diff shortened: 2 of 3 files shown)


Daniel15:
Thanks!

PR #726, opened 2026-07-14
Propagate testOptions through snapshot/inline test helpers (fixes #581)

### LINKED ISSUE #581: testOptions aren't propagated through some of the test methods
`applyTransform` relies on `testOptions` for a `parser` option.
At the same time `defineSnapshotTestFromFixture`, `defineSnapshotTest`, `runSnapshotTest` (maybe others) don't propagate user-supplied `testOptions`.

A possible work-around is to run export the parser const from the transformer module, but that limit the ability to use the module with multiple transformers (e.g `babylon` and `ts`)

LeSingh1:
## Problem

`applyTransform` selects the transform's parser from `testOptions.parser`, but several public `testUtils` helpers never forward `testOptions`, so the option is silently ignored (issue #581):

- `runSnapshotTest` and `defineSnapshotTest` drop `testOptions` entirely.
- `defineInlineTest` does not accept it.
- `defineSnapshotTestFromFixture` is self-inconsistent: it *accepts* `testOptions` and uses `testOptions.parser` to pick the fixture file extension (e.g. `foo.input.tsx`), but then runs the transform with the **default** parser. A TypeScript fixture is therefore read as `.tsx` yet parsed with babel, throwing on TS-only syntax.

Only `runInlineTest`/`runTest` currently honor `testOptions.parser`. As noted in #581, the documented work-around (exporting `parser` from the transform module) forces a single parser per module and doesn't compose.

## Fix

Thread `testOptions` through the affected helpers as a **trailing optional argument**, matching the existing `runInlineTest`/`runTest` signatures:

- `runSnapshotTest(module, options, input, testOptions)`
- `defineInlineTest(module, options, input, expectedOutput, testName, testOptions)`
- `defineSnapshotTest(module, options, input, testName, testOptions)`
- `defineSnapshotTestFromFixture(...)` now forwards its already-accepted `testOptions` to `defineSnapshotTest`.

All changes are backward compatible (new optional parameters only; downstream defaults to `{}`).

The README is updated to document the `testOptions` argument for these helpers.

## Verification

Added a `testOptions propagation` describe block plus a `test-tsx-transform` fixture whose input contains an `enum` (accepted only by the `tsx`/`ts` parsers). Each helper is exercised with `{ parser: 'tsx' }`.

- Against `main`, 4 of the 5 new cases fail with `Unexpected reserved word 'enum'` (only the already-correct `runInlineTest` passes), demonstrating the bug.
- With the fix, all pass.

```
$ npx jest src/__tests__/testUtils-test.js --ci
Tests:       21 passed, 21 total
Snapshots:   9 passed, 9 total

$ npx jest   # full suite
Test Suites: 18 passed, 18 total
Tests:       1 skipped, 213 passed, 214 total
```

Fixes #581

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,6 +511,9 @@ const transformOptions = {};
 defineInlineTest(transform, transformOptions, 'input', 'expected output', 'test name (optional)');
 ```
 
+An optional `testOptions` object can be passed as the last argument to select the
+parser used to run the transform, e.g. `defineInlineTest(transform, transformOptions, 'input', 'expected output', 'test name', { parser: 'tsx' })`.
+
 #### `defineSnapshotTest`
 
 Similar to `defineInlineTest` but instead of requiring an output value, it uses Jest's `toMatchSnapshot()`
@@ -521,6 +524,8 @@ const transform = require('../myTransform');
 const transformOptions = {};
 defineSnapshotTest(transform, transformOptions, 'input', 'test name (optional)');
 ```
+
+Like `defineInlineTest`, an optional `testOptions` object (e.g. `{ parser: 'tsx' }`) can be passed as the last argument to select the parser.
 
 For more information on snapshots, check out [Jest's docs](https://jestjs.io/docs/en/snapshot-testing)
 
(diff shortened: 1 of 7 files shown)


Daniel15:
Thanks!

